PR #282, opened 2015-11-19
Add support for xcode 7.1+ (osx 10.11)

### LINKED ISSUE #277: 'openssl/evp.h' file not found issue when build project with xcode 7.0
hi, when build project with xcode 7.0, compile issue 'openssl/evp.h' file not found in SUDSAVerifier.m.

Any update for this ?

lrusak:
I originally wanted to add support for openvpn git version, but I ran across this issue building tunnelblick on OSX 10.11 (I see now that openvpn git version has been added :+1: )

This fixes #277 

The changes work like this
 - Openvpn is built before sparkle, this is so that openssl can be built and statically linked in sparkle.
 - ~~Headers from the openssl build are symlinked to the sparkle source directory~~
 - ~~The headers are then updated to reflect this in the SUDSAVerifier.m file~~
 - A patch is used to adjust the sparkle xcode project to include the built openssl headers and libs, this is then dynamically adjusted depending on your machine. This patch also links sparkle against libcrypto and libz.

~~There were some whitespace issues (because of using tabs) and this was problematic while using atom editor. I can get rid of this if you desire.~~

Using this I get a working build on my 10.11 system. I'd like to know if this works on your system and older versions of xcode.

Let me know if you want me to make any changes or feel free to pick the commits directly and edit them to your desire.

~~EDIT: I may have spoke too soon. TunnelBlick runs but whenever I try to add an ovpn file I get this~~
>An error (status 190) ocurred while trying to check the security of the lukas-macbook configuration. Please quit and relaunch Tunnelblick. If the problem persists, please reinstall Tunnelblick

~~Not sure if this is just a problem with master or a problem with this PR, I will have to build a vanilla master build to test. Here is a log, http://pastebin.com/raw.php?i=xtq7ERjr~~

~~Seems to be an issue here,~~ 
```
2015-11-18 23:51:46 Tunnelblick[32615] tunnelblickd status from compareShadowCopy: 190
                                       tunnelblickd stderr:
                                       'Must be a directory: /Users/lukas/Library/Application Support/Tunnelblick/Configurations/lukas-macbook.tblk'
2015-11-18 23:51:46 Tunnelblick[32615] Internal Tunnelblick error: unknown status 190 from compareShadowCopy(lukas-macbook)
2015-11-18 23:51:57 Tunnelblick[32615] An uncaught exception was raised
2015-11-18 23:51:57 Tunnelblick[32615] must provide a launch path
```
So it looks like it doesn't have to do with this PR

### Update:

I have updated this PR, the symlinking and stuff is absolutely not needed when using the openssl include and lib dirs in the sparkle patch I created. It is much more simple now and produces a working build for me. The original PR commit can be seen here, https://github.com/lrusak/Tunnelblick/commit/f7aa72d4038c08a65131a35d8f5269b661f7848d

### DIFF
--- a/third_party/Makefile-common
+++ b/third_party/Makefile-common
@@ -486,7 +486,12 @@ built-sparkle-prepare:
 	@echo "THIRD_PARTY: Patching Sparkle.framework..."
 	if [ -d $(PATCHES_DIR)/sparkle ]; then \
 		cd "$(SPARKLE_BUILD_DIR)"; \
-		for patch_file in `ls $(PATCHES_DIR)/sparkle/*.diff`; do \
+		if [ $(XCODE_VERSION_MAJOR) -ge 0700 ] ; then \
+			patches="`ls $(PATCHES_DIR)/sparkle/*.diff`" ; \
+		else \
+			patches="`ls $(PATCHES_DIR)/sparkle/[0-9][0-9]*.diff`" ; \
+		fi ; \
+		for patch_file in $${patches}; do \
 			patch_name="$$(basename "$${patch_file}")" ; \
 			patch -p1 -N --dry-run -i $$patch_file > /dev/null; \
 			if [ $$? == 0 ]; then \
@@ -500,6 +505,9 @@ built-sparkle-prepare:
 				echo "error: Sparkle patch could not be applied: $(SPARKLE_NAME)/$${patch_name}" ; \
 			fi \
 		done; \
+		sed -i '' -e "s|@OPENSSL_STAGING_DIR@|$(OPENSSL_STAGING_DIR)|g" \
+			  -e "s|@TARGET_ARCH@|$(CURRENT_ARCH)|g" \
+			     "$(SPARKLE_BUILD_DIR)/Extras/Source Code/Sparkle.xcodeproj/project.pbxproj" ; \
 		cp -f -p -X "$(PATCHES_DIR)/sparkle/de.lproj/Sparkle.strings" "Extras/Source Code/de.lproj/Sparkle.strings" ; \
 		cp -f -p -X "$(PATCHES_DIR)/sparkle/it.lproj/Sparkle.strings" "Extras/Source Code/it.lproj/Sparkle.strings" ; \
 		cp -f -p -X "$(PATCHES_DIR)/sparkle/ru.lproj/Sparkle.strings" "Extras/Source Code/ru.lproj/Sparkle.strings" ; \
--- a/third_party/sources/patches/sparkle/add-default-includes-and-libs.diff
+++ b/third_party/sources/patches/sparkle/add-default-includes-and-libs.diff
@@ -1,0 +1,42 @@
+--- a/Extras/Source Code/Sparkle.xcodeproj/project.pbxproj	2008-09-13 20:55:00.000000000 -0700
++++ b/Extras/Source Code/Sparkle.xcodeproj/project.pbxproj	2015-11-18 17:00:30.000000000 -0800
+@@ -976,6 +983,12 @@
+ 			isa = XCBuildConfiguration;
+ 			baseConfigurationReference = FA1941CA0D94A70100DD942E /* ConfigFrameworkDebug.xcconfig */;
+ 			buildSettings = {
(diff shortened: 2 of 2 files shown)


jkbullard:
Thank you very much. I will examine at this in detail in the next few days, but the general approach looks good.

I especially appreciate your extensive description of the patch.

NOTE: The problem you're having with adding a configuration is due to a regression in commit e6764837e90a499fc6a62a6fae47a9bd005477be (the huge "Build 4430" commit.) I will commit a fix for that in a few minutes (assuming all goes well).

jkbullard:
OK, I have committed a fix; you should be able to add configurations now after you apply the fix.

Sorry about the that.

lrusak:
I have updated the PR and included additional information in the original message.

I have simplified things as the symlinking is not needed when using the patch for sparkle to include the openssl include and lib dirs.

My only concern now, and I have no way of testing this, is what happens on old versions of xcode where openssl is built for multiple arches, I'm not sure if sparkle will need to be linked against all the different archs for openssl or how that would work. Perhaps I can just add a switch to test for new xcode?
```
if [ $(XCODE_VERSION_MAJOR) = 0700 ] ; then \
    do stuff
else
    don't do anything
fi
```

jkbullard:
First, thanks again for working on this. Building on Xcode 7+  is important to many people.

1. Building OpenVPN before Sparkle works with Xcode 3.2.2 and multiple architectures, so **I have just made commit 93cf9c1b087819f72a5ab774b756f74bfff49c5d to do that**. (The reason Sparkle was built first is that when I was debugging Sparkle patches I wanted builds to fail earlier, before undergoing the long process of building OpenVPN.)

2. This patch makes the Xcode 7+ builds fragile: if Sparkle's project.pbxproj file changes significantly, the .diff will break. But that's acceptable to me; I don't anticipate such changes until a more modern version of Sparkle replaces the existing one, and have no time for that to be done soon. And presumbably such a newer version of Sparkle will itself deal with this problem (probably breaking on Xcode 3.2.2, but by then that will probably be OK and we will abandon PPC and Tiger).

3. I don't think it will work with multiple architectures. The "sed" edits will only be done for the first architecture. Maybe instead of ending up with "$(OPENSSL_STAGING_DIR)/i386" (or whatever), it could end up as "$(OPENSSL_STAGING_DIR)/XXX" where "XXX" returns an Xcode variable that contains the architecture being built at the time. I don't know what Xcode variable "XXX" would be, and I don't know if the Xcode architecture names are the same as the names that OpenSSL uses but I have a vague memory that they are _not_ the same, so that may not work, or it might be messier.

An alternative would be to enclose everything in "IF Xcode 7" blocks, but that won't work because Tunnelblick's patching mechanism does not have the ability to patch only under specified conditions, so it can't be done only for Xcode 7+.

But I think this could be done relatively easily by

1. Moving the "06-add-default-includes-and-libs.diff" file to be /third_party/add-default-includes-and-libs-to-sparkle.diff, so it is outside of the regular patching mechanism.

2. Adding code in Makefile-common that looks like the following:

    if [ $(XCODE_VERSION_MAJOR) -ge 0700 ] ; then  \
        Patch Sparkle.xcodeproj/project.pbxproj with the .diff
        Use "sed" to make the other changes to the file
    fi

(Note: The "-ge 0700" is so it will work for Xcode 8, 9, etc. when they are released).

That way, the changes only happen for Xcode 7 and higher, and building using Xcode 3.2.2 is unchanged.

Again, thanks for all your work on this. It's getting close!

lrusak:
I made the changes that were requested. 

Some minor notes:
 - I used ```ls $(PATCHES_DIR)/sparkle/[0-9][0-9]*.diff``` so that the patch could still reside in the patches directory (to keep things all together).
 - I changed the patch slightly to work with the openssl staging directories

If you would like me to add some checks to the patch like you do with the other patches (like using --dry-run and checking for an exit status) just let me know and I can put them in as well.

Yes the patch is fragile, feel free to ping me if it ever breaks. I am commited now :+1:

jkbullard:
Great -- thanks. I like your idea and way of keeping the patch in the Sparkle patch directory.

If you could add the checks to the patch (--dry-run, checking for exit status), I can get this into the main branch.

lrusak:
Updated, I made it so there is the least amount of extra lines possible and no redundant code.

The sed line always runs, but it is harmless if there is nothing to replace.

PR #283, opened 2015-11-19
Add support for lz4 to openvpn

lrusak:
Not sure if you were already working on support for this :) I was doing snappy as well then realized that snappy support was dropped from openvpn.

The change is pretty simple and lzo was used as a template:
 - lz4's default install only installs the binaries, so a simple ```cp``` is needed to install the lib and header
 - Changes to openvpn don't effect the older openvpn versions as it just skips lz4.
 - The lz4 package was received from https://github.com/Cyan4973/lz4/archive/r131.tar.gz then renamed

```
OpenVPN 2.3.8 x86_64-apple-darwin [SSL (OpenSSL)] [LZO] [PKCS11] [MH] [IPv6] built on Nov 19 2015
```
```
OpenVPN 2.3_git_7546cba x86_64-apple-darwin [SSL (OpenSSL)] [LZO] [LZ4] [PKCS11] [MH] [IPv6] built on Nov 19 2015
```
If I missed anything let me know :+1:


lrusak:
Rebased and updated the way it builds, it now uses the specific makefile in the libs dir to build and install the static binary and headers.

jkbullard:
Thanks! It looks pretty straightforward. My only concern about including it is that it is "one more thing" to keep track of -- for updates, security vulnerabilities, etc. (That's assuming the memory footprint is small if it is not used, which I *think* is true.)

The original OpenVPN commit that includes lz4 support says
<blockquote>
LZ4 does not compress as well as LZO or Snappy, but needs far less CPU
and is much faster, thus better suited for mobile devices.
</blockquote>
 
So why should it go into Tunnelblick, which is only on non-mobile devices with (generally) plenty of CPU power? Is the idea to have Tunnelblick be the VPN server for mobile devices?

Or, in general, can you describe **why** lz4 should be included in Tunnelblick? Do typical OpenVPN setups (in Linux distributions; on Windows) support it?

lrusak:
Your concerns are noted, and I'll try to address them.

1. As far as I have tested it is only possible to support a single compression method at a time. So if the server supports lz4 then only that can be used, if it only support lzo then only that can be used (I am not 100% sure on this, however, I see errors in my logs that seem to support this theory). So on my openVPN server on my VPS I use lz4 because I mainly use it with my iphone.

2. No I don't think many OpenVPN servers support it at the moment, but I imagine adoption will be wider when OpenVPN 2.4 is finally released.

I currently build my own OpenVPN version for use on my vps, see [here](https://copr.fedoraproject.org/coprs/lrusak/openvpn-git/)

So in summary, I can't imagine this would be widely used at the moment, but in the future I'm sure it will be. It's for you to decide if it's worth supporting it sooner or later.

PS. snappy support has been removed from OpenVPN git

jkbullard:
<blockquote>As far as I have tested it is only possible to support a single compression method at a time. So if the server supports lz4 then only that can be used, if it only support lzo then only that can be used (I am not 100% sure on this, however, I see errors in my logs that seem to support this theory).</blockquote>

You might want to look into whether you can use a client-side config file to choose which compression method to use. The openvpn.8 doc says:

<blockquote>
In a server mode setup, it is possible to selectively turn
compression on or off for individual clients.

First, make sure the client-side config file enables selective
compression by having at least one
"--comp\-lzo"
directive, such as
"--comp\-lzo no".
This will turn off compression by default,
but allow a future directive push from the server to
dynamically change the
on/off/adaptive setting.

Next in a
--client\-config\-dir
file, specify the compression setting for the client,
for example:
"comp\-lzo yes"
"push "comp\-lzo yes""

The first line sets the
comp\-lzo
setting for the server
side of the link, the second sets the client side.
</blockquote>

Although the above discusses turning lzo compression on/off, it might work for choosing which method of compression to use with the "--compress" option. (The info in this doc often lags behind what is implemented.)

Another option would be to create two server instances of OpenVPN: one instance using LZO and the other instance using LZ4.

Finally, the "git-master" version of OpenVPN included in Tunnelblick 3.6beta16 supports LZ4 but does not statically link to an lz4 library. So I think if the lz4 **command** (which you would need to add yourself) is available on your Mac, LZ4 will "just work" by using the command instead of a library. (That's assuming it is similar to OpenSSL: if OpenVPN is built without the OpenSSL libraries, it uses the openssl command. But I don't *know* if the same is true for LZ4.)

**All that aside**, if your patch "works for me", I will probably merge it. It may be a while, though, before I can do the necessary testing. Thanks again for offering it.

jkbullard:
Thanks! I've merged it.

PR #293, opened 2016-01-03
I made brand new pull request regarding travis-ci integration

chipitsine:
https://travis-ci.org/chipitsine/Tunnelblick-travis/builds/99934582

only .travis.yml added (it uses xcode 7 now)


jkbullard:
Thanks for your work on this.

Please notice the following on the [Travis CI Getting Started page](https://docs.travis-ci.com/user/getting-started/):
<blockquote>
Note: You can only enable Travis CI builds for repositories you have admin access to.
</blockquote>

I have decided not to merge this pull request into the official Tunnelblick repository. I do not wish use Travis CI for the official project, and because Travis CI requires admin access to the GitHub repository, anyone who wishes to use it will have to use it on a repository to which they have admin access -- that is, their own repository. Because they would then be building from their own repository, they can simply add the .yml file to their repository themselves.

==========

But I can see that this file would be useful for someone willing to use Travis CI, so I would be happy to add it to the [User Contributions section of the Tunnelblick Downloads page](https://tunnelblick.net/downloads.html#user-contributions). If I do so, I would include something like the following as the description:

<blockquote>
This file allows users to create unofficial Tunnelblick builds on [Travis CI](https://docs.travis-ci.com). Such builds are not official Tunnelblick builds and do not run on PowerPC processors nor on all of the versions of OS X on which the official builds run. In addition, the builds are not digitally signed, which means that certain parts of Tunnelblick will not work: Gatekeeper in its default setting will not allow the builds to be installed, the program's use of the OS X Keychain will be restricted if such a build is replaced, and the TUN and TAP drivers will not load into recent versions of OS X. Finally, the use of a third-party (Travis CI) to build Tunnelblick means that the third party must be trusted.
</blockquote>

If you would like me to do this, please let me know here, and please include any changes/comments to the above and the name or pseudonym I should give as the contributor.

chipitsine:
that page on travis-ci lead you to misunderstanding, it means that you must have admin rights, not travis, here's description on what travis actually has (and you confirm it when adding repo to travis): https://docs.travis-ci.com/user/github-oauth-scopes/

as for the rest, I do mind to add it as "contributed" docs, however, I'm afraid, it will get broken soon if it is not involved in daily use. You have removed xcode 6.3 support, and it got broken already (I fixed it today), also it is not very usefull right now, it just allows to see whether build was green or red. I am going to add cache (to avoid rebuilding openssl each time), code signing and customization.

ok, let us place it somewhere, I would say that github wiki would be better, but google code wiki is also ok.

you are a bit wro g about "creating builds", travis-ci does not keep artefacts, one needs to upload them, for example, to s3 amazon. right now it just allows you to see build status (either green or red) and place a badge to github. and to automatycally build pull requests, no more yet

as I will build signed customized release, I will let you know:)

travis is good, it provides you with mac os x build agents (we have no mac os x agents in our teamcity)

jkbullard:
To repeat: I have no use for Travis CI. Using it would not help me because I have a build infrastructure that I am happy with and suits my needs as the Tunnelblick developer. In addition, I do not want to authorize their website to have *any* access the official Tunnelblick repository on GitHub.

(For the paranoid: actually, the "real" official repository is hosted on my own equipment and only I have access to it; GitHub is a publishing mechanism for the source code and a distribution mechanism for binaries.)

I'm not saying Travis CI is not "good"; I am saying that I personally do not need or want to use it.

If and when Travis CI creates builds for other people, they need to understand the limitations of those builds. And because of the way Travis CI works, they need to create those builds on their own repositories. So they do not need the .yml file to be in the official Tunnelblick repository, they only need it to be in their own repository.

As to your concern that a "User Contributed" .yml file would not be kept updated, I would not maintain the .yml file anyway, so it wouldn't be updated by me even if it were in the Tunnelblick source code.

If you do not want to make it "User Contributed", I suggest you create your own GitHub project for this (Note that Google Code has been read-only for several months.)

PR #303, opened 2016-04-23
Update Localizable.strings ZH_CN

linux-mac:
Update some wrong words: ZH_CN


jkbullard:
Thanks for your work, but we don't accept pull requests for localization.

Please see [Localizing and Translating Tunnelblick](https://tunnelblick.net/cLocalizeTranslate.html).

PR #313, opened 2016-07-21
Add GNU General Public License, version 2 licence file

mitchellcash:
In accordance with most other repositories it is probably a good idea to add a copy of the GNU General Public License, version 2 to the root of the repository.

Also updated the tunnelblick/COPYING file to the latest "plain text" version of the GPLv2.


jkbullard:
Great! Thank you!

PR #314, opened 2016-07-21
Update README

mitchellcash:
Content essentially remains the same but gave it a small facelift.


jkbullard:
The facelift looks great; thanks!

PR #318, opened 2016-07-27
Tunnelblick OpenVPN Xor Patch URL correction

iamGavinJ:


jkbullard:
Thanks!

Looking at this, I noticed some outdated info, which I will fix in a few minutes.

PR #319, opened 2016-07-28
pt-br translation errors

eduardodaluz:
pt-br translation errors

### DIFF
--- a/tunnelblick/PT_BR.lproj/Localizable.strings
+++ b/tunnelblick/PT_BR.lproj/Localizable.strings
@@ -1,8 +1,8 @@
 /* Status message */
-"Disconnect All (%@)" = "Desconetar Todas (%@)";
+"Disconnect All (%@)" = "Desconectar Todas (%@)";
 
 /* Status message */
-"Disconnect All (%d Connections)" = "Desconetar Todas (%d Conexões)";
+"Disconnect All (%d Connections)" = "Desconectar Todas (%d Conexões)";
 
 /* Status message */
 "No Active Connections" = "Nenhuma Conexão Ativa";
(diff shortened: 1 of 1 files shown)


jkbullard:
Thanks, we appreciate all the work you have put into this. However:

1. We don't do localization changes in the source code directly; we do translations in Crowdin and then have the translations merged into the source code from there. To join the localization team, please [contact the developers](https://tunnelblick.net/reachus.html).

2. According to our Portuguese and Portuguese (Brazilian) translators:

"Conectar" is deprecated in Brazil and Portugal, but not only that word ("conectar")... ALL word were there's a "c" before the "t" must have the "c" removed, like so:

conectar —> conetar
facturar —> faturar
facto —> fato
afecto —> afeto
acto —> ato
actor —> ator

ALSO words with hyphen in the middle ("-") must remove the hyphen, like so:

conectar-se —> conetarse
desconectar-se —> desconetarse 

This orthographic agreement officially called "Portuguese Language Orthographic Agreement of 1990" was made in 1990 but only in 2004 was implemented, furthermore Portuguese countries had from 2004 to 2011 to make the change and pass the new and CORRECTED way of spelling to others, like in Schools. Its expansion goes far beyond only Portugal and Brazil, including also other countries that speak Portuguese like Angola, Cape Verde, Guinea Bissau, Galicia, etc.

Here are some references online of this agreement. The first page it's in Portuguese, the second page it's the same link but to the section with some examples, the third link it's a sporadic (bad) translations of the same agreement.

References:
https://pt.wikipedia.org/wiki/Acordo_Ortográfico_de_1990
https://pt.wikipedia.org/wiki/Acordo_Ortográfico_de_1990#Mudan.C3.A7as_nos_restantes_pa.C3.ADses_lus.C3.B3fonos
https://en.wikipedia.org/wiki/Portuguese_Language_Orthographic_Agreement_of_1990

Again, thanks for making the effort, and please do let us know of any other problems you see. (In the translations or anywhere else in the program.)

Best regards,

Jon Bullard

eduardodaluz:
ok

Issue #317, opened 2016-07-25
FEATURE REQUEST [in-progress]: OpenVPN binary with LibreSSL


iamGavinJ:
x-linking to Google Groups:
https://groups.google.com/d/msg/tunnelblick-discuss/ijYitWvgG0Y/P7VviHHaCwAJ

Thanks.

jkbullard:
Tunnelblick 3.6.9beta01 includes versions of OpenVPN that are static-linked with LibreSSL. The user can select OpenSSL or LibreSSL versions on the "Settings" tab.

Many thanks to @iamGavinJ for his encouragement and PRs.

PR #320, opened 2016-07-28
Phase 1 of third-party easy-updating

iamGavinJ:
@jkbullard, pushing this over to you now just as a precaution in case I do something silly with my repo.  You don't need to commit it unless you really want to.  The script is fully functional up untill the point of retrieving the latest openvpn/master, pushing it into the correct folder and copying over patches and configure-options from the existing directory.

I'm working on automating these bits for openvpn to use it as a template before doing the work on LibreSSL. #317


jkbullard:
@iamGavinJ, Thank you, but without knowing what the next phases are, I don't understand how this fits into Tunnelblick. I assume they result in automating updates of the third-party software that Tunnelblick uses, and for reasons described below, I am reluctant to do that.

I will leave this PR open in case I misunderstand its purpose or an explanation of the rest of the phases you plan causes me to change my mind.

I do not want to include this particular functionality as part of the Tunnelblick build process, and although the script would be useful for me as the Tunnelblick maintainer, I have my own scripts to perform this and most of the other tasks involved in maintaining Tunnelblick. (I have not released them publicly because I don't want to have to document them properly or answer questions about using them.)

Tunnelblick's build process (what happens when you "Build" in Xcode) does not include the automatic downloading of source code (or anything else) from the Internet. There are too many security issues with doing that, and Tunnelblick's patches of third-party sources (including, for example, [patching OpenVPN to add the "scramble" option](https://tunnelblick.net/cOpenvpn_xorpatch.html)) can be broken by updates to the source code. In addition, the OpenVPN "git-master" is not always in a "good" state for inclusion into Tunnelblick. It is rare, but sometimes commits are made that break things and additional commits are then required to un-break them.

Tunnelblick uses a small number of third-party software items, but they are updated infrequently and it is (in my view) much safer for me to keep track of their updates, make adjustments to our patches if necessary, test them for compatibility, and then include them in new releases. That process is usually quick; security updates to OpenSSL, for example, are usually available in Tunnelblick within 24 hours of being available; often much sooner than that.

PR #326, opened 2016-09-05
Added third_party temp file and products ignores

iamGavinJ:


jkbullard:
Thanks, but this makes several **other** changes to the Makefile in addition to moving it. I want to be able to focus on such changes separately.

Please redo this as three PRs:
 1. .gitignore changes
 2. Moving Makefile and symlinking to it
 3. Changes to the Makefile

(And you'll get credit for three PRs accepted instead of one, too! : )

iamGavinJ:
I'm not sure how that second commit made it into the PR as it was supposed to be one commit per PR. It is also duplicated in PR #327 for that reason. 

I'll reverse out the commit. 

I'm having to create a separate branch for each PR because github doesn't allow the specific bundling of commits into a PR, just a pointer to the head of that branch, so it's not exactly an elegant, straightforward process.

jkbullard:
Ah, sorry it is so painful, I can see your point. But I want to make changes as transparent as possible, so it is important.

(Actually, using git is painful in itself in several ways, but it is great in others.)

iamGavinJ:
It's no problem. Leave it with me. I'll take another look at it tomorrow.

iamGavinJ:
Jon (@jkbullard) this PR has been corrected now by hard resetting the commit to the previous one so you should be right to accept it now.

jkbullard:
👍 ; thanks.

PR #328, opened 2016-09-05
Renamed the OpenSSL Makefile so that it's not accidently called or included/sourced

iamGavinJ:


jkbullard:
I don't understand why this is needed. If it is to avoid problems from typos in a command-line, Tunnelblick isn't built from the command line.

Why would only this file need to be renamed -- wouldn't the other makefiles have the same problem?

iamGavinJ:
Because SSL was the primary reason for the set of changes I started the makefile-allssl from scratch to replace makefile-openssl.  All other dependent packages and their makefiles were updated in place with the existing make targets stubbed out by prefixing them with 'zzz'

jkbullard:
I'm sorry, but I still don't understand this PR.

It renames "Makefile-openssl" to "~Makefile-openssl". Why do you want to do that?

PR #330, opened 2016-09-05
This Makefile replaces Makefile-openssl.

iamGavinJ:
The design iterates over any file found in the sources directory
with a filespec regex of `.*/(libre|open){1}ssl-.*tar.gz$` and extracts and
builds it.

If you wish to exclude or include versions of SSL simply save them
into the sources directory and dependent versions of pkcs11-helper
and openvpn will iterate over the extracted versions building statically
against these.

Patches are located in third_party/sources/patches/ssl/ again prefixed
with the flavour of SSL.


jkbullard:
@iamGavinJ - Thanks for doing this.

The next commit, 3a79bb73c9420ea7fa7e7ee1e089f22a5a6eaa01, makes changes including changes to Makefile-allssl so this works under Xcode when building Tunnelblick.

PR #338, opened 2016-10-08
Increase KEY_SIZE to safe default of 2048

adamtbeames:
Default vars are using a KEY_SIZE=1024 which is no longer considered safe.
Public CA's have moved to a keysize of 2048 (Browsers have also removed 1024 bit certs!)

This changes defaults in all locations I could find to use KEY_SIZE=2048

### DIFF
--- a/third_party/sources/easy-rsa/easy-rsa-2-tunnelblick/vars
+++ b/third_party/sources/easy-rsa/easy-rsa-2-tunnelblick/vars
@@ -52,7 +52,7 @@ export PKCS11_PIN="dummy"
 # down TLS negotiation performance
 # as well as the one-time DH parms
 # generation process.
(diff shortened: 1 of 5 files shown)


jkbullard:
The only place it should be changed is in third_party/sources/easy-rsa/easy-rsa-2-tunnelblick/vars. That is the only place it is used by Tunnelblick.

The other references are in "easy-rsa-from-openvpn-2.2.1". That is a copy of what was distributed with OpenVPN 2.2.1, so it should not be changed; that folder can be used to create a .diff to show what changes were made to create "easy-rsa-2-tunnelblick".

adamtbeames:
Thanks for the info.
I've reverted changes to all files except for the relevant one.

jkbullard:
Great - thanks!

PR #346, opened 2016-12-08
Update german Localizable.strings

joerg-wille:
wrong german translation


jkbullard:
We do translations on Crowdin and export into the Tunnelblick source code, so no PRs are accepted. A different translation has been accepted in Crowdin.

PR #369, opened 2017-04-11
keychain management client support for openvpn >= 2.4.0

kaloprominat:
hi!

since version 2.4.0 openvpn supports offloading of encryption/decryption operations onto macOS keychain via management interface. It becomes possible by adding to openvpn new configuration option "--management-external-cert" which specifies identity search template for management program.

more details about this feature can be found here:

http://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg09717.html
https://sourceforge.net/p/openvpn/mailman/openvpn-devel/thread/20141208115202.GA4807%40cachalot/#msg33125844
https://sourceforge.net/p/openvpn/mailman/openvpn-devel/thread/20150112085121.GA25193%40cachalot/#msg33225603

In order it to work properly, management program should answer to management commands NEED-CERTIFICATE and RSA_SIGN. 

This patch adds some cryptographic modules for working with certificates/keychain related items, and two new reactions on commands "NEED-CERTIFICATE" and "RSA_SIGN" to VPNConnection.m. With this patch, it makes possible to use keychain stored certificates/keys for openvpn authentication.

Code based on keychain management client daemon example, contributed to openvpn as en example for this feature: https://github.com/OpenVPN/openvpn/tree/master/contrib/keychain-mcd.

As the result, you may specify openvpn config like

>...
>client
>tls-client
>management-external-cert "macosx-keychain:subject:CN=your certificate cn"
>management-external-key

and Tunnelblick asks keychain access and provides valid signature to openvpn while connecting.

This also allows to use same way certificates/keys, stored on some external token, which is shown as own keychain in keychain access.

Successfuly tested with rutoken: https://www.rutoken.ru/products/all/rutoken-ecp/

**UPD:**
Successfuly tested with Aladdin-rd eToken (SafeNet eToken 5100)

### DIFF
--- a/tunnelblick/Tunnelblick.xcodeproj/project.pbxproj
+++ b/tunnelblick/Tunnelblick.xcodeproj/project.pbxproj
@@ -241,6 +241,11 @@
 		B8C9FB49170C77A400BCA243 /* LeftNavViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = B8C9FB46170C77A400BCA243 /* LeftNavViewController.m */; };
 		B8D807031832EEC400A0B937 /* NSTimer+TB.m in Sources */ = {isa = PBXBuildFile; fileRef = B8D807021832EEC400A0B937 /* NSTimer+TB.m */; };
 		B8F6264C1641C8C700943C59 /* easyRsa.m in Sources */ = {isa = PBXBuildFile; fileRef = B8F6264B1641C8C700943C59 /* easyRsa.m */; };
+		C09E285C1EAE86C700F07A74 /* base64.c in Sources */ = {isa = PBXBuildFile; fileRef = C09E28521EAE86C700F07A74 /* base64.c */; };
+		C09E285D1EAE86C700F07A74 /* cert_data.c in Sources */ = {isa = PBXBuildFile; fileRef = C09E28541EAE86C700F07A74 /* cert_data.c */; };
+		C09E285E1EAE86C700F07A74 /* common_osx.c in Sources */ = {isa = PBXBuildFile; fileRef = C09E28561EAE86C700F07A74 /* common_osx.c */; };
+		C09E285F1EAE86C700F07A74 /* crypto_osx.c in Sources */ = {isa = PBXBuildFile; fileRef = C09E28581EAE86C700F07A74 /* crypto_osx.c */; };
+		C09E28601EAE86C700F07A74 /* keychain_mcd.c in Sources */ = {isa = PBXBuildFile; fileRef = C09E285A1EAE86C700F07A74 /* keychain_mcd.c */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -638,6 +643,16 @@
 		B8D807021832EEC400A0B937 /* NSTimer+TB.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = "NSTimer+TB.m"; sourceTree = "<group>"; };
 		B8F6264A1641C8C700943C59 /* easyRsa.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = easyRsa.h; sourceTree = "<group>"; };
 		B8F6264B1641C8C700943C59 /* easyRsa.m */ = {isa = PBXFileReference; explicitFileType = sourcecode.c.objc; fileEncoding = 4; path = easyRsa.m; sourceTree = "<group>"; };
+		C09E28521EAE86C700F07A74 /* base64.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = base64.c; sourceTree = "<group>"; };
+		C09E28531EAE86C700F07A74 /* base64.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = base64.h; sourceTree = "<group>"; };
+		C09E28541EAE86C700F07A74 /* cert_data.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = cert_data.c; sourceTree = "<group>"; };
+		C09E28551EAE86C700F07A74 /* cert_data.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = cert_data.h; sourceTree = "<group>"; };
+		C09E28561EAE86C700F07A74 /* common_osx.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = common_osx.c; sourceTree = "<group>"; };
+		C09E28571EAE86C700F07A74 /* common_osx.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = common_osx.h; sourceTree = "<group>"; };
+		C09E28581EAE86C700F07A74 /* crypto_osx.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = crypto_osx.c; sourceTree = "<group>"; };
+		C09E28591EAE86C700F07A74 /* crypto_osx.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = crypto_osx.h; sourceTree = "<group>"; };
+		C09E285A1EAE86C700F07A74 /* keychain_mcd.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = keychain_mcd.c; sourceTree = "<group>"; };
+		C09E285B1EAE86C700F07A74 /* keychain_mcd.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = keychain_mcd.h; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
@@ -728,6 +743,7 @@
 		080E96DDFE201D6D7F000001 /* Classes */ = {
 			isa = PBXGroup;
 			children = (
+				C09E28511EAE869A00F07A74 /* Keychain-mcd */,
 				B8407882170CE4E10002B941 /* Authorization & Keychain */,
 				B840787F170CE4130002B941 /* Common */,
 				B8407881170CE4B00002B941 /* Class Extensions & Subclasses */,
@@ -1333,6 +1349,23 @@
 			name = Uninstaller;
 			sourceTree = "<group>";
 		};
+		C09E28511EAE869A00F07A74 /* Keychain-mcd */ = {
+			isa = PBXGroup;
+			children = (
+				C09E28521EAE86C700F07A74 /* base64.c */,
+				C09E28531EAE86C700F07A74 /* base64.h */,
+				C09E28541EAE86C700F07A74 /* cert_data.c */,
+				C09E28551EAE86C700F07A74 /* cert_data.h */,
+				C09E28561EAE86C700F07A74 /* common_osx.c */,
+				C09E28571EAE86C700F07A74 /* common_osx.h */,
+				C09E28581EAE86C700F07A74 /* crypto_osx.c */,
+				C09E28591EAE86C700F07A74 /* crypto_osx.h */,
+				C09E285A1EAE86C700F07A74 /* keychain_mcd.c */,
+				C09E285B1EAE86C700F07A74 /* keychain_mcd.h */,
+			);
+			name = "Keychain-mcd";
+			sourceTree = "<group>";
+		};
 /* End PBXGroup section */
 
 /* Begin PBXLegacyTarget section */
@@ -1793,6 +1826,7 @@
 				633DF4591312B13100B8606C /* ConfigurationManager.m in Sources */,
 				639A1100132AC9D800754C2A /* StatusWindowController.m in Sources */,
 				63ABFE9C1349E435003DEB64 /* LoginWindowController.m in Sources */,
+				C09E285C1EAE86C700F07A74 /* base64.c in Sources */,
 				6388563E134B7E1C008DF8C1 /* PassphraseWindowController.m in Sources */,
 				635ECCF51378885600F67F8D /* DBPrefsWindowController.m in Sources */,
 				635ECCFA137888A800F67F8D /* MyPrefsWindowController.m in Sources */,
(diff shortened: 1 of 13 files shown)


kaloprominat:
@jkbullard, what is your thoughts for approving such feature? there is some details in implementation, which i'd like to discuss

tunnelblick/Tunnelblick.xcodeproj/project.pbxproj
```diff
@@ -2058,6 +2096,7 @@
 				GCC_WARN_UNUSED_PARAMETER = YES;
 				GCC_WARN_UNUSED_VALUE = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
+				HEADER_SEARCH_PATHS = "../third_party/build/openssl-1.0.2k/include/";
```
kaloprominat: some modules require openssl features, so i added build openssl path as headers search path. i'm not sure, is there a better way to link to openssl?

jkbullard:
@kaloprominat - Thanks!

I like the idea of adding this feature, but I won't be able to fully review this patch for at least a week or two (or possibly more).

I do have two initial questions:
 1. Does this require OpenSSL, or will it also work with LibreSSL?
 2. Could it be reworked to use Apple's built-in crypto instead of OpenSSL?

kaloprominat:
That was really reasonable questions, because it turned out, that previous openssl link was missed and totally unnecessary at all. Now patch doesn't depend on any ssl, except Apple's, and no additional header search paths needed.

jkbullard:
Thanks for doing all this work.

I've taken an initial look at the PR. I haven't looked "line by line" at the code yet, but so far I have the following observations/comments:

**I like the basic idea as I understand it:** to extend the existing code in Tunnelblick's that handles the OpenVPN "management interface" so that it also processes the "NEED-CERTIFICATE" and "RSA_SIGN" commands.

**Comment One**

 1. Six new files (cert_data.h&c, common_osx.h&c, and crypto_osx.h&c are similar to files in the OpenVPN 2.4.1 source "contrib/keychain-mcd" subfolder.

 1. Two new files (base64.h&c) are similar to files in the 2.4.1 source "src/openvpn" subfolder.

I assume that for simplicity you copied/pasted code from the existing files into new files in the Tunnelblick source instead of somehow linking to the files that already exist in the OpenVPN 2.4.1 source. That bothers me because it will be difficult to maintain. I would prefer something that would "link" to the existing files. I admit that may be difficult or impossible:
 * Tunnelblick usually includes multiple versions of OpenVPN source; which version should be used?
 * Can Xcode handle symlinks in a way that could work for this?

I will play around with this idea and see what I can come up with.

**Comment Two**

All of the new files have copyright notices of
`Copyright © 2017 Pavel Kondratiev. All rights reserved.`
(They were apparently generated by Xcode automatically for you.)

But the files are based on existing files in the OpenVPN source code as described above. Those have the following copyright notice:
`Copyright (C) 2010 Brian Raderman <brian@irregularexpression.org>`
` Copyright (C) 2013-2015 Vasily Kulikov <segoon@openwall.com>`
(Vasily Kulikov was the author of the patch that was committed to OpenVPN master branch as 39e3d336d4eeab847a3395ddeb430e0a9ca387b9 on Wed Feb 25 19:07:18 2015 +0300.

============
I'm not asking for any changes to your PR (especially because if we can link to the existing source files, eight of the ten files in the PR will go away), but I am interested in your thoughts on what I've written so far.

kaloprominat:
@jkbullard, thank you for reviewing PR.

Yeah, i have a few thoughts about your comments.

**Thought one**

As for similar files from openvpn sources, although it very similar to original, but it still seems that original sources have a couple of minor issues. For example, сert_data.c includes unneccecary openssl link, which i wrote about: `#include <openssl/ssl.h>`. It doesn't used anywhere, just missed there by a mistake i guess. Also, copied version of sources fixes a couple of potential memory leaks issues. Speaking strictly, i assume that keychain-mcd part, included in openvpn sources, is not intended for reviewing/testing and maintaining by openvpn developers, neither for including in openvpn builds. In my opinion, it should be considered as a referal implementation on management-side, or as an example. Besides, some part of original implementation located in main.c file, which can't be reused the way you suggested. So if a possible change could be done, that affects both main.c and cert_data/common_osx/crypto_osx parts, it becames that original files could not be used in tunnelblick no more. Considering all that thoughts and potential problems of linking to original files, i would preffer to use an adopted copy of it's implementation, either of struggling with linking to it. Imho, it would be not less reliable as an original copy, because of original copy is not even maintained by any openvpn devs, i assume.

What about base64.h&c, as for me, it's pretty straightforward implementation of simple two encoding/decoding base64 functions, and no more. It's simple, common and clear enought. It can be copied, reworked or implemented in other module own way. But as for me, i would be very unusual to link to theese functions, while it's located in sources of other product.

**Thought two**

I'm newbie to licensing and all this stuff, but i may guess, copyrights need to be kept. As far as i know, if openvpn is licensed under GNU GPLv3, and i should left "copyleft" of other authors notices, am i right? Could you, please, give me a little guide of how to make this code rights-eligible and respectful to other autors rights? Thank you!



All those is only my thoughs, as i'm not enforcing any decision of how this patch should be implemented and accepted in Tunnelblick. I will agree to any changes you ask, becase i want this feature to be presented in tunnelblick much more, rather than the way of how this patch should be implemented.

jkbullard:
@kaloprominat - Thanks for your comments/clarifications. (And for the patch, too, of course!)

Given what you wrote, copying files from the OpenVPN source into Tunnelblick as you did makes sense. However, I would like to make it clear where the files come from and what changes have been made to them.

I'm not a licensing expert either, but I think the following should work to address my concerns:

 1. Start with a fresh copy of the Tunnelblick source from the master branch.
 2. Add unmodified copies of the eight nearly-the-same-as-OpenVPN files from the (expanded) source code that is in third_party/openvpn/OpenVPN 2.4.1/openvpn/. (That is, cert_data.h&c, common_osx.h&c, crypto_osx.h&c, and base64.h&c.). (Don't add them to the project yet, just copy them into the "tunnelblick" folder.)
 3. Commit with a message like "Add copies of support files for keychain management from OpenVPN 2.4.1 source" and an extended comment (or whatever git calls it) that references https://sourceforge.net/p/openvpn/mailman/openvpn-devel/thread/20150225160718.GA6306%40cachalot
(which shows the full history of what was finally committed to OpenVPN).
 4. Insert the standard Tunnelblick copyright notice at the start of each of the files (with the existing OpenVPN copyright notice immediately after it). You can find the standard notice at the start of VPNConnection.h. Just replace the existing two copyright lines with a line like
`Modifications copyright 2017 Pavel Kondratiev. All rights reserved.`
 5. Commit with a message like "Add Tunnelblick copyright notices to OpenVPN's support files for keychain management"
 6. Make other changes to the files as needed (for example, removing `#include <openssl/ssl.h>`
 7. Commit with a message like "Modify OpenVPN's support files for keychain management" and 
 8. Add the two new files (keychain_mcd.h&c) and make changes to VPNConnection.h&m) and Tunnelblick.xcodeproj. Add some comments to keychain_mcd.h&c to describe where they came from -- as I understand it, they are mostly or all routines from the main.c file.
 10. Commit with a message like "Hook up support for keychain management"

The way to add the files to the project repository without adding them to the Xcode project is to just copy the files in Finder. To add the new files to the project (change Tunnelblick.xcodeproj) In step 8 you can drag/drop the files from a Finder window onto the file list in Xcode.

This would make it clear where the code is coming from and what changes were made to it and clearly separates the copyright changes from the actual code changes. It will also make it much easer for me to review the code line-by-line, after which I may have further changes to suggest.

Thanks again!

kaloprominat:
@jkbullard thank you for help! I followed your guide and would like you to take a look on this branch compared to Tunnelblick master: https://github.com/Tunnelblick/Tunnelblick/compare/master...kaloprominat:keychain-mcd-pr

If everything is fine, i could replace commits in this PR with those. Or i will do whatever you suggest as better.

jkbullard:
@kaloprominat - Yes, sorry, I should have been more specific. I did mean for you to replace the Tunnelblick copyright notices copied from VPNConnection.h with the "Modifications…".

If you don't want to do the entire process over again, just make another commit to restore the OpenVPN copyrights and replace the Tunnelblick copyrights.

kaloprominat:
That's ok, i can handle git.
There is correct copyright replacement version: https://github.com/Tunnelblick/Tunnelblick/compare/master...kaloprominat:keychain-mcd-pr

the same branch comparison

jkbullard:
Thanks. I am (slowly) working my way through and testing the code and your changes and will reply here when I have something to say; I hope it will only be a couple more days.

(I'm looking at https://github.com/Tunnelblick/Tunnelblick/compare/master...kaloprominat:keychain-mcd-pr)

jkbullard:
@kaloprominat - A complication: The OpenVPN developers are considering removing keychain-mcd from the source code. See https://sourceforge.net/p/openvpn/mailman/openvpn-devel/thread/b9a70fef-feda-053c-bd96-c2b0e74c576a%40sf.lists.topphemmelig.net/#msg35824609.

kaloprominat:
@jkbullard - well, and what's your thoughts on this? They taking mostly, as i understand it, about memory allocations checks. It sounds like a not a big deal, i guess.

jkbullard:
@kaloprominat - I have just [responded](https://sourceforge.net/p/openvpn/mailman/message/35825217/) to the OpenVPN developers.

Apparently the OpenVPN developers are looking for someone to maintain keychain-mcd and fix several problems that have been discovered. The original author of the code isn't responding to their emails, so they are asking if anyone else is using the code and is willing to take over responsibility for it.

That isn't something that I want to do, but would you consider doing that?

kaloprominat:
@jkbullard i may try. But, let me think a while. I actually have contact of original keychain-mcd author, as i've had some conversations with him a few years ago. I would try to contact him and ask his point on this. If he won't be able to help, i'll consider how i can do it by myself.

But, anyway, is maintaining keychain-mcd in openvpn - a mandatory for this patch to be apply? I mean, the openvpn devs are talking about removal of contrib/keychain-mcd example implementation itself, not the openvpn part, that support crypto functions offloading. Patch will be also usable without keychain-mcd part in openvpn.

What you think?

jkbullard:
I like the functionality this PR provides and am willing to accept it even if keychain-mcd is being removed from OpenVPN.

But I don't want to introduce bugs, and apparently the OpenVPN developers have found some bugs. It isn't clear to me if the problems they found are in parts of keychain-mcd that your PR doesn't use, but even if they are, they could be fixed in the Tunnelblick implementation.

Life has intervened, and I haven't begun my review of the patch yet. Because it is in such a security-sensitive area and bugs have already been found -- by you and apparently by others -- I intend to be extra careful in the review, so it will take longer than I originally expected.

dsommers:
We have received some code audit responses to the keychain-mcd code which is quite concerning.  And as this code is code which others build up-on, it needs to be improved.  The overall coding style is in my eyes quite poor and ambitious.

To be honest, I think it makes sense to me that the keychain-mcd code is moved to Tunnelblick.  This is more or less the most commonly used open source OpenVPN package for macOS.  And keychain-mcd is specifically targetting macOS.

I am willing to help out in such a transition phase.

Regarding the licensing (I've only quickly skimmed through here, I might have missed a few details).  The OpenVPN code is GPLv2 only (we don't have the _or later version_ part in the license texts).  That includes the keychain-mcd code.  So as long as you keep the copyright and license as is and the project incorporating this project is GPLv2 only compatible, there should be no licensing issues.   For some more input, this is a pretty good and comprehensive overview:
https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#Good_Licenses

kaloprominat:
@dsommers , as i'm not very experienced C coder, so could you help us and point us to an obviously poor code parts of keychain-mcd? You wrote about memory allocations checks – did you mean something, like, that every malloc() should be followed by, for example, malloc_size() check?

dsommers:
@kaloprominat That's no problem, I can help out reviewing changes.

The issues discovered so far is that there are almost no checks on what `malloc()` or most of the `CF*()` function calls returns.  On success, they most commonly return a pointer (a memory address) to the memory buffer allocated.  But on errors, they return `NULL`.  `NULL` is an address which is invalid, so if the application tries to write data to an address set to `NULL`, it will in the best scenarios just crash.

So the first step is actually to check if all functions used which can return `NULL` gets a proper error check.  Right now, the code may very well attempt to write to a `NULL` address if an error happened, which will fail badly.

Take this code fragment from `cert_data.c`:
```C
    CertDataRef pCertData = (CertDataRef)malloc(sizeof(CertData));
    pCertData->subject = GetFieldsFromCertificate(certificate, kSecOIDX509V1SubjectName);
```

A pointer to a structure called `CertDataRef` is being allocated, and the pointer to this structure is stored in `pCertData`.  As long as `pCertData` is not `NULL`, everything is okay and it is possible to access the struct member `subject` like what is done in the line above: `pCertData->subject`.

However, if `malloc()` failed, it would return `NULL` instead.  Then `pCertData` would also point at the address `0`.  And that is defined as illegal.  So when you then try to use  `pCertData->subject`, the program triggers a so-called segmentation fault (aka segfault or `SEGV`).

The proper thing to do is then to add some error checks:
```C
    CertDataRef pCertData = (CertDataRef)malloc(sizeof(CertData));

    if (NULL == pCertData)
    {
        /* do some error handling */
        return;  /* or return NULL, if that is more appropriate */
    }
    pCertData->subject = GetFieldsFromCertificate(certificate, kSecOIDX509V1SubjectName);
```

This error handling doesn't need to be complex.  Some times you don't need to do anything else than just return/exit the function.  Other times you want to add an error message of what went wrong.  Whether you need to just exit the function via `return;` or `return NULL;` depends on what this function is declared to return.  Some places functions may also return `false` or `0` in error situations; this needs to be carefully analyzed in each function separately.

This is basically the first step which needs to be done.  Evaluate all function calls which allocates memory, like most of the `CF*` functions and all `malloc()`/`calloc()`, and add proper error checking.  The next step is to evaluate other possible failure scenarios which is not accounted for in the code; are there other functions which may return a type of error signal which is not being acted upon?

A starting point for the Apple documentation on the `CFArray` functions can be found here:
https://developer.apple.com/reference/corefoundation/cfarray-s28


You may have spotted that the error check is `if (NULL == pCertData)` and not `if (pCertData == NULL)`.  Both works, but the latter one will cause a bug if there's a typo and you use `=` instead of `==`.  Using the former approach, the compiler will barf at you - you can't assign a value to `NULL`.  So using the former approach is recommended and being a more defensive coding style.

kaloprominat:
@dsommers Okay, I get the idea, thank you for clarifying. I will start looking and improving all such cases in keychain-mcd code and later provide another version for review. May i count on your further assitance with review?

dsommers:
Absolutely! I can definitely help out on the generic C code stuff.  Unfortunately, when it comes to macOS specifics, I have no more clue than what the documentation says; I don't have access to any macOS based environments. But I hope others with access to macOS can go clarify things there if it is needed.

dsommers:
Any progress on this issue?  I'm would like to propose to remove this code from the upstream OpenVPN code base soon.

And also get a confirmation if this code will be moved over to the Tunnelblick project instead would be helpful.

Just trying to tighten up loose threads.

kaloprominat:
Unfortunately, I had not enough time to start work on this yet due to vacation season. But i definitely will as soon as i have possibility. Something near to august, i guess.

I'm not sure, who's in charge to confirm any keychain-mcd movement. But as i guess, @jkbullard won't be tend to accept keychain-mcd in Tunnelblick "as is", with all those issues this thread about.

@dsommers, you are talking about exactly only contrib/keychain-mcd part removal, are you? The openvpn management interface support for keychain is not meant to be removed, right?

jkbullard:
@dsommers - confirming the assumptions made by @kaloprominat:

- Yes, I will not accept keychain-mcd into Tunnelblick "as is"; and

- Yes, I will accept it into Tunnelblick if/when the issues are fixed.

dsommers:
@kaloprominat Yes, I am only talking about removing the contrib/keychain-mcd part.  The management interface in OpenVPN will be unmodified.

I will then submit a patch to upstream OpenVPN to remove this code from their codebase and point at this ticket for further information.

chalam-volety:
@kaloprominat thank you for your work towards this feature. I have run into this need and am wondering if you've had any chance to take it further.

sascha-eglau-dragon:
@kaloprominat have you given up on this PR? We would like to use this feature :)

hardlydavis:
Noting interest myself here too. Would be great to see this implemented in Tunnelblick or OpenVPN. Sadly I lack the C knowledge to help myself.

boss-chifra:
This is critical functionality, to store the certificate in a non-clear text file.

kaloprominat:
Actually in my organization we've been using this patch with Tunnelblick on thousands of macs for 7 years already and haven't seen any problems, which we could identified as related to memory-leaks or something like that.

I hope i will have time soon to update PR for latest Tunnelblick version support.

mpdude:
Is my understanding correct that without this patch, private keys have to be put in plain text into the configuration file?

PR #374, opened 2017-05-19
Add tls-crypt option

EmTeedee:
Adds the new tls-crypt option available in OpenVPN 2.4

I don't have a Mac myself, so can't test it, sorry.


jkbullard:
Thanks!

PR #391, opened 2017-08-11
Build doc: Changed legacy build direction to use project-specific setting.

Bo98:
Instead of directing people to change global Xcode settings which will affect their existing projects, it may be better to use the project-specific setting which works just as well.


jkbullard:
Thanks.

PR #392, opened 2017-08-11
Updated automake and libtool

Bo98:
I'm not sure why automake 1.9 is still in the version check, but I replaced 1.15 with 1.15.1.


jkbullard:
Thanks!

Issue #386, opened 2017-06-22
Replace OpenSSL 1.0.2k with 1.0.2l and LibreSSL 2.5.0 with 2.5.4


l2dy:
LibreSSL 2.5.1 fixes CVE-2016-7056.

Issue #353, opened 2017-01-20
Weird problems running on Sierra if build with Xcode 8


jkbullard:
If Tunnelblick is built with Xcode 8 on Sierra (including at least up through 8.2.1), it has weird problems when running on Sierra. For example:
 1. After connecting, the icon shows the connected status correctly, but if you click on the icon the status is described as "No Active Connections".

 2. Tunnelblick neither logs that the IP address changed (when it should), nor complains that the IP address didn't change. This may be related to 1. 

The problems happen consistently when running on Sierra but don't happen when running on El Capitan.

Bo98:
From my tests on Xcode 8.3.3 on macOS 10.12.6:

1. I do not experience this issue.
2. I experience the opposite - it complains that the IP didn't change **except** when "route all IPv4 traffic through the VPN" is enabled where it detects the IP change correctly.

jkbullard:
@Bo98 - Thanks for your report, which seems to indicate that it works properly for you.

I will continue build the official Tunnelblick releases using Xcode 7 on El Capitan until I get reliable results (and at this point I will probably skip to Xcode 9).

Bo98:
Considering Xcode 9 will be releasing next month, that seems fair.

jkbullard:
Closing this -- Xcode 9 seems to fix this problem.

PR #393, opened 2017-08-11
Updated third party libraries

Bo98:
* OpenSSL 1.0.2k -> 1.0.2l (#386)
* LibreSSL 2.5.0 -> 2.5.5 (#386)
* LZO 2.09 -> 2.10
* LZ4 r131 -> 1.7.5
* pkcs11-helper 1.11 -> 1.22
* easyrsa 3.0.0-rc2 -> 3.0.1

I did not update Sparkle since I would not be able to test it properly and see if the upgrade process still works with the complex patches.


DomT4:
Did you manage to connect to networks using the updated LibreSSL? I started working on something similar in my [own fork](https://github.com/DomT4/Tunnelblick/commits/third_party_tweaks) but encountered major connection issues on macOS 10.12 when using either LibreSSL 2.5.4 or 2.5.5 that I hadn't been able to pin down yet.

Bo98:
I was able to connect to my VPN server fine. I've not noticed any issues with LibreSSL vs OpenSSL so far using Xcode 8.3.3 on macOS 10.12.6.

I could send you a copy built from my machine if you want to try it out.

jkbullard:
I have a private test branch with LibreSSL 2.5.5 that doesn't have any problems connecting on 10.11.6.( (I haven't tested it on 10.12 yet.)

@DomT4 - Can you post logs or other info about the problems you had connecting?

DomT4:
I'll pull this PR locally later today & build it. Maybe there's something amiss with what I've been doing locally in testing my own changes. I'm running 10.13/Xcode 9 now which isn't wildly helpful but at least build errors are broadly more obvious than runtime errors if any are encountered.

jkbullard:
@DomT4 - There may be problems with Xcode 9: see #353 which describes some problems when running a binary built with Xcode 8; and, of course, 10.13 beta could also be a problem.

DomT4:
Apologies for the delay from me here. Everything seems to work fine with this PR & indeed Xcode 9, so colour me bemused but relieved. Would say this is good to go from my point of view.

There's an unrelated long-term issue that if you leave `p11-kit` linked through Homebrew it semi-silently breaks the build in a fairly mysterious way unless you actually go log hunting, but that's not relevant to this PR.

jkbullard:
This works for me as well. Thanks!

PR #394, opened 2017-08-11
Made use of native highlighting for menu icon

Bo98:
Now supports Yosemite-style highlighting.

This replaces the "highlighted-" mechanism that was only used for the current standard icon set.

I do not have a copy of OS X previous to Yosemite to test this, but the API is not new.


jkbullard:
Thank you!

PR #395, opened 2017-08-11
Switched preference icons to use those supplied by the OS (supports Retina display and Yosemite styling)

Bo98:
Before: 

After: 

(it will use older-style images pre-10.10, but I do not have a installation to test how it looks)

This change does **not** include the lock icon. There is not a standard lock icon image accessible _directly_ via public API. So the options are:
* Create a custom icon.
* Copy Apple's icon from SecurityInterface.framework. (May need two versions to fit both pre-10.10 and Yosemite styling, along with Retina@2x versions)
* Use `SFAuthorizationView`. This however isn't for use in the toolbar and it bypasses your `SystemAuth` class.
* Use private API `SFAnimatedLockButton`. It happens to also be animated. A proof of concept I tried just as a fun experiment is at https://github.com/Bo98/Tunnelblick/commit/744fe3815e022b87d88dd38c90d5fd2b7ac35ead. I didn't include it in this pull request since it is private API, but could do so if this was the preferred approach.
 

---

This pull request also replaces the add, remove and action menu button images used in the Configurations pane in preferences.


Bo98:
Sorry about that, it should work now.

What's your opinion on the lock icon options?

jkbullard:
I really like using the system images, and they work fine on 10.7.5.

But the Info panel doesn't scroll automatically (on 10.11, anyway), and the .xibs generate several warnings when the project is built.

Maybe modifying a .xcodeproj or .xib with Xcode 8 changes it too much for it to be used in Xcode 7 afterwards.

@Bo98, Can you redo this, using Xcode 7? That might help isolate the problems.

jkbullard:
For the lock image, I think we should either:
 * Use a custom icon as we do now -- Retina-compatible and updated to look modern (I don't care about the look on pre-10.10 but it would be nice to have that, too);
or
 * Use ```SFAuthorizationView```. I'm not sure what problems that could cause, so it would have to be done carefully.

Private APIs are a big pain.

Bo98:
> But the Info panel doesn't scroll automatically (on 10.11, anyway)

That's odd, considering this PR doesn't touch the info panel XIB. I can't reproduce on 10.12, but I do have a copy of 10.11 that I will try.

> the .xibs generate several warnings when the project is built.

Apple aren't very good at making collaboration-friendly XIBs it seems. I'll download Xcode 7 on 10.11 and try it out.

> Maybe modifying a .xcodeproj or .xib with Xcode 8 changes it too much for it to be used in Xcode 7 afterwards.

.xcodeproj doesn't have any Xcode 8 changes btw - the changes are removal of files.

---

This is what `SFAuthorizationView` looks like btw (from System Preferences):

Bo98:
I've now redone the XIBs in Xcode 7.

I'll try out `SFAuthorizationView` in a separate pull request.

jkbullard:
Sorry, but that doesn't do it -- see the screenshot of warnings below. I think the only way to do it is to use Xcode 7 to modify the original xibs (which haven't been touched by Xcode 8).

I'm guessing that the warnings have to do with replacing the add/remove/action buttons.

I suggest you create a new PR that only replaces the large buttons. That would accomplish most of the visual improvement and wouldn't require any mods to the .xibs, and is more granular (do as I say, not as I do : )

This would mean leaving the image file for the small info button as it is. Those small info-buttons will disappear anyway if/when we implement mouseover help for the controls that need it -- currently we only do it for checkboxes and simple buttons, I think, but not for drop-down buttons.


screen shot 2017-08-18 at 4 17 11 pm

Bo98:
Those are the exact warnings that should be fixed now. I completely redid the XIB in Xcode 7 so they haven't been touched by Xcode 8 at all in the latest commit.

Bo98:
Just did a clean build with the latest commit and those warnings definitely don't happen for me anymore in Xcode 7.

jkbullard:
Sorry, that was my mistake; I didn't properly get the new commit correctly and was using an old branch.

Looks good now, thanks a lot!

PR #396, opened 2017-08-11
Allow manual scrolling of credits

Bo98:
Last change - sorry for the influx.

Manual scrolling wasn't stopped before completely, but it fought with the auto-scroll and was very jittery. Now it temporarily pauses the auto-scroll for at least one second.


jkbullard:
@Bo98 - **Thank you very much** for the "influx"!

I am away for the weekend and won't be able to deal with these PRs for a few days but will do as soon as I can.

jkbullard:
Thanks!

PR #399, opened 2017-08-16
ShellScriptToInstallAutotools: minor script tweaks

DomT4:
Each commit is explained in the commit message.

### DIFF
--- a/third_party/ShellScriptToInstallAutotools.sh
+++ b/third_party/ShellScriptToInstallAutotools.sh
@@ -46,6 +46,14 @@ if [ ! -d "${downloads_folder_path}" ] ; then
     exit 1
 fi
 
+for buildtool in automake autoconf libtool glibtool; do
+  if [ -L /usr/local/bin/$buildtool ]; then
+    echo "/usr/local/bin/$buildtool exists & appears to be a symlink."
+    echo "Presuming $buildtool is from a package manager & refusing to overwrite."
+    exit 1
+  fi
+done
+
 echo "INSTALLING AUTOCONF:"
 cd "${downloads_folder_path}"
 curl -OL "${autoconf_url}"
--- a/third_party/ShellScriptToInstallAutotools.sh
+++ b/third_party/ShellScriptToInstallAutotools.sh
@@ -74,7 +74,7 @@ cd "${downloads_folder_path}"
 curl -OL "${libtool_url}"
 tar xzf "${libtool_version}.tar.gz"
 cd "${libtool_version}"
-./configure
+./configure --program-prefix=g
 make
 sudo make install
 


DomT4:
Feel free to voice opinion here but if you like it don't merge it just yet. I'm going to tweak the second commit just slightly later today to cover a potentially valid use case.

jkbullard:
@DomT4 - Thanks!
 1. Do the build scripts that Tunnelblick.xcodeproj magically know to use glibtool, or would they need to be modified? The reason for this script is that the Apple-supplied tools are not present or not sufficient; shadowing the Apple libtool is OK with me but I see why it might not be OK for everyone.
 2. If you don't use Homebrew, the script can't be used to update. Can we detect Homebrew and only exit if it is installed? Or ask the user if they want to replace the existing installation(s)?

DomT4:
> Do the build scripts that Tunnelblick.xcodeproj magically know to use glibtool

As far as I'm aware from repeated local builds all of the third-party projects TB pulls in know to check for `glibtool`. Because both MacPorts and Homebrew have been installing as that for a long time to avoid shadowing Apple's `libtool` it's become quite a popular thing to check for.

> If you don't use Homebrew, the script can't be used to update. 

Aye, this is the potentially valid use case I mentioned above. I was thinking of tweaking the check to only bail if those files exist _and_ are symlinks. Thoughts? It keeps it fairly simple at least & I can't imagine why you'd only symlink into `/usr/local/bin` if you weren't using Homebrew there. Otherwise, if you'd prefer, I can go hunting for `brew` & check if those executables exist and bail if they do. Whichever sounds like a better idea to you.

jkbullard:
> As far as I'm aware from repeated local builds all of the third-party projects TB pulls in know to check for glibtool. Because both MacPorts and Homebrew have been installing as that for a long time to avoid shadowing Apple's libtool it's become quite a popular thing to check for.

OK, thanks. According to [this Stack Overflow answer](https://stackoverflow.com/questions/22677133/installation-of-libtool), we should also rename "libtoolize" to "glibtoolize", but [Apple itself](https://developer.apple.com/library/content/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html#//apple_ref/doc/uid/TP40002850-SW5) only mentions glibtool.

> I was thinking of tweaking the check to only bail if those files exist and are symlinks. Thoughts? 

Yes, I think that's less fragile than looking for "brew".

DomT4:
> According to this Stack Overflow answer, we should also rename "libtoolize" to "glibtoolize"

Aye, this should be done here as the `program-prefix` argument applies to everything.

> Yes, I think that's less fragile than looking for "brew".

I've flipped the test flag to `-L` rather than `-f` which should do the trick & seems to work fine from local testing.

DomT4:
Not sure if with the change to `-L` whether you want me to refine the `echo` a little more, to state explicitly that we're checking for symlinks rather than simply refusing to overwrite in every scenario. I've changed the commit message but obviously not everyone is gonna bother to read commit messages, for better or worse.

jkbullard:
> Aye, this should be done here as the program-prefix argument applies to everything.
OK, thanks. I didn't consider that.

> Not sure if with the change to -L whether you want me to refine the echo a little more, to state explicitly that we're checking for symlinks rather than simply refusing to overwrite in every scenario.

Yes, please.

DomT4:
Language tweaked.

jkbullard:
Thanks!

PR #400, opened 2017-08-18
Fixed linking error when GnuTLS, NSS or mbed TLS is found

Bo98:
The linking error comes when linking pkcs11-helper to OpenVPN (although it shouldn't really be happening since it's built statically).

pkcs11-helper will always prefer OpenSSL, which we are always using anyway (LibreSSL is compatible with OpenSSL), but it would still always try to link with GnuTLS, NNS and mbed TLS if they were found. Therefore we can safely disable using them, and disabling them fixes issues caused by them being linked.


jkbullard:
Thanks!

PR #426, opened 2018-01-17
Check for lack of IP addr before assuming DHCP

### LINKED ISSUE #402: Logic for detection of DHCP gateway under TAP is not accurate (client.up.tunnelblick.sh)
````	
	bRouteGatewayIsDhcp="false"
	if [ -z "${route_vpn_gateway}" -o "$route_vpn_gateway" == "dhcp" -o "$route_vpn_gateway" == "DHCP" ]; then
		bRouteGatewayIsDhcp="true"
	fi
````

Why do we assume that it's a DHCP gateway in the absence of that environment variable? `-z "${route_vpn_gateway}"`  

This breaks under my config, which is a pretty basic LAN only TAP config where I want to set the nameserver.  A static IP and DNS IP is pushed from the OpenVPN server, but Tunnelblick incorrectly thinks it needs to use DHCP due to the absence of a "route_vpn_gateway".  Well, in this config there is no gateway being used! It's LAN-only.

mpsrig:
Fixes #402 

We should not assume DHCP just because of a lack of "route_vpn_gateway" env var, this is an open issue on openvpn:  https://community.openvpn.net/openvpn/ticket/668

It seems a safe bet to not use DHCP if there is already an IP address configured (i.e. by openvpn).

Please see further discussion on the issue.  Thanks!

### DIFF
--- a/tunnelblick/client.up.tunnelblick.sh
+++ b/tunnelblick/client.up.tunnelblick.sh
@@ -1471,8 +1471,17 @@ if ${ARG_TAP} ; then
 	# Still need to do: Look for route-gateway dhcp (TAP isn't always DHCP)
 	bRouteGatewayIsDhcp="false"
 	if [ -z "${route_vpn_gateway}" -o "$route_vpn_gateway" == "dhcp" -o "$route_vpn_gateway" == "DHCP" ]; then
-		bRouteGatewayIsDhcp="true"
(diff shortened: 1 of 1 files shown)


mpsrig:
@jkbullard

jkbullard:
@mpsrig  - Looks good. Thank you!

PR #433, opened 2018-02-04
public-dns-list: add free and public dns in China

JLHwung:
Added free and public dns in China. Here is a list of their websites (in Chinese):

114dns: http://www.114dns.com/about.html
alidns: http://www.alidns.com/
dnspod: https://www.dnspod.cn/products/public.dns
baiduDNS: https://dudns.baidu.com/intro/publicdns/

This list is not covered by http://pcsupport.about.com/od/tipstricks/a/free-public-dns-servers.htm, but most China users will use these public DNS in pushed network configurations. So it is confused to warn user while they do configure to use `114.114.114.114`.

> DNS servers '114.114.114.114' will be used for DNS queries when the VPN is active
> NOTE: The DNS servers do not include any free public DNS servers known to Tunnelblick. This may cause DNS queries to fail or be intercepted or falsified even if they are directed through the VPN. Specify only known public DNS servers or DNS servers located on the VPN network to avoid such problems.

### DIFF
--- a/tunnelblick/FreePublicDnsServersList.txt
+++ b/tunnelblick/FreePublicDnsServersList.txt
@@ -22,4 +22,7 @@ Hurricane Electric	74.82.42.42
 puntCAT	109.69.8.51	 
 Yandex 77.88.8.8  77.88.8.1
 Global Cyber Alliance	9.9.9.9
-
+114dns	114.114.114.114	114.114.115.115
+alidns	223.5.5.5	223.6.6.6
+dnspod	119.29.29.29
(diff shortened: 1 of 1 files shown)


jkbullard:
Thank you very much, especially for providing the justification and links.

jkbullard:
I've updated this list from lifewire.com and sorted it to make it easier to modify in the future, in 
bc7865dd2d939de035878f8923d23423700d1830 and  d6546998eae51c63058f6f3adc9e4095742735d1.

Thanks again, @JLHwung, for the nudge.

PR #437, opened 2018-02-28
Fix typo in regex in client.up.tunnelblick.sh

cvgs:
It looks like there is an error in the regex. Currently it references the ".app" directory inside "/Applications/Tunnelblick". Changing "/" to "\\" should fix that.


jkbullard:
@cvgs - Thanks!

Note: This only affects a Tunnelblick.app that isn't in /Applications -- only "Debug" builds used internally by developers, not any builds that are released to users.

PR #439, opened 2018-03-03
Improve hungarian translations

szdavid92:
Improve hungarian translation by
  - uniformizing some commonly used phrases
  - using the official names for macOS terms such as Desktop or Keychain
  - correcting misinterpretations
  - keeping easy-rsa as is

### DIFF
--- a/tunnelblick/HU.lproj/Localizable.strings
+++ b/tunnelblick/HU.lproj/Localizable.strings
@@ -86,7 +86,7 @@
 "Delete Credentials" = "Hitelesítőadatok törlése";
 
 /* Button */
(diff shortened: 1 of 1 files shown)


jkbullard:
@szdavid92  - I'm sorry, but we don't accept PRs for translations.

We do translations on Crowdin.com, where they can be voted up or down by other translators. Then we import translations with the highest vote totals from Crowdin into the Tunnelblick source code. It's a one-way transfer from Crowdin into the source code – we don't transfer from the source code into Crowdin, so we don't accept PRs.

Please see [Localizing and Translating Tunnelblick](https://tunnelblick.net/cLocalizeTranslate.html).

Email developers@tunnelblick.net to volunteer to translate.

PR #464, opened 2018-06-23
 Handle OpenVPN Challenge/Response static protocol

faucct:
Rebased https://github.com/Tunnelblick/Tunnelblick/pull/430, but left off the OTP generation feature. I have failed to test it because of blocked System Extensions. Can someone confirm it works or tell me how to make it work when building from source?

### DIFF
--- a/tunnelblick/AuthAgent.m
+++ b/tunnelblick/AuthAgent.m
@@ -295,7 +295,7 @@ TBSYNTHESIZE_NONOBJECT_GET( BOOL,       showingPassphraseWindow)
                     }
                     [passwordKeychain deletePassword];
                     if (  [passwordKeychain setPassword: passwordLocal] != 0  ) {
-                        NSLog(@"Could not save password in Keychain '%@'", [usernameKeychain description]);
+                        NSLog(@"Could not save password in Keychain '%@'", [passwordKeychain description]);
                     }
                     [gTbDefaults setBool: YES forKey: usernameAndPasswordPreferenceKey];
                     [gTbDefaults removeObjectForKey:  usernamePreferenceKey];
--- a/tunnelblick/MF_Base64Additions.h
+++ b/tunnelblick/MF_Base64Additions.h
@@ -1,0 +1,34 @@
+//
+//  MF_Base64Additions.h
+//  Base64 -- RFC 4648 compatible implementation
+//  see http://www.ietf.org/rfc/rfc4648.txt for more details
+//
+//  Designed to be compiled with Automatic Reference Counting
+//
+//  Created by Dave Poirier on 2012-06-14.
+//  Public Domain
+//  Hosted at https://github.com/ekscrypto/Base64
+//
+
+#import <Foundation/Foundation.h>
+
+@interface NSString (Base64Addition)
++(NSString *)stringFromBase64String:(NSString *)base64String;
++(NSString *)stringFromBase64UrlEncodedString:(NSString *)base64UrlEncodedString;
+-(NSString *)base64String;
+-(NSString *)base64UrlEncodedString;
+@end
+
+@interface NSData (Base64Addition)
++(NSData *)dataWithBase64String:(NSString *)base64String;
++(NSData *)dataWithBase64UrlEncodedString:(NSString *)base64UrlEncodedString;
+-(NSString *)base64String;
+-(NSString *)base64UrlEncodedString;
+@end
+
+@interface MF_Base64Codec : NSObject 
++(NSData *)dataFromBase64String:(NSString *)base64String;
++(NSString *)base64StringFromData:(NSData *)data;
++(NSString *)base64UrlEncodedStringFromBase64String:(NSString *)base64String;
++(NSString *)base64StringFromBase64UrlEncodedString:(NSString *)base64UrlEncodedString;
+@end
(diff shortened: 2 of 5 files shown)


jkbullard:
@faucct - Thanks.

This PR creates a separate dialog after the dialog that asks for the username/password. Although that would work, I have two problems with it:
 1. It doesn't comply with the description in the [Management Notes](https://github.com/OpenVPN/openvpn/blob/master/doc/management-notes.txt):
> The static protocol differs from the dynamic protocol in that the challenge question and response field is given to the user in the initial username/password dialog…
 2. It is "yet another dialog" inserted into the flow.

It might be better to pass this to the existing username/password dialog so it can be included as a third field. But I'm not clear on how this gets used – maybe it would be better as a separate dialog, at least sometimes – for example, if the username/password are stored in the Keychain.

**Before proceeding further on this, I'd like to hear comments about the above from folks who use or want to use static challenge.**

Here are a few comments about the changes to VPNConnetion.m (after a very preliminary look):
 1. The NSAssert at line 3470 is not needed; the situation it checks for is handled two lines later.
 2. `echoResponse` is not implemented, it is just ignored. Offhand I'm not sure it can be done with an `NSSecureTextField`, but if not, an `NSTextField` could be used instead if `echoResponse` is TRUE.
 3. The code to get `echoResponseStr` and `challengeText` seems unnecessarily convoluted since they are at fixed offsets from rngStartChallenge. Something like the following seems simpler (to me, anyway : ) but I wouldn't insist on it:
```
NSString * echoResponseStr = [line substringWithRange:NSMakeRange(rngStartChallenge + [@" SC:" length], 0)];
NSUinteger challengeStart = rngStartChallenge + [@" SC:X," length];
NSString * challengeText = [line substringWithRange:NSMakeRange(challengeStart, [line length] - challengeStart];
```
 4. I prefer less complicated if/else statements. Instead of
```
if (!response) {
    [self sendStringToManagementSocket:[NSString stringWithFormat:@"password \"Auth\" \"%@\"\r\n", escaped(myPassword)] encoding:NSUTF8StringEncoding];
} else {
    [self sendStringToManagementSocket:[NSString stringWithFormat:@"password \"Auth\" \"SCRV1:%@:%@\"\r\n", [myPassword base64String], [response base64String]] encoding:NSUTF8StringEncoding];
}
```
I'd rather see:
```
if (response) {
    [self sendStringToManagementSocket:[NSString stringWithFormat:@"password \"Auth\" \"SCRV1:%@:%@\"\r\n", [myPassword base64String], [response base64String]] encoding:NSUTF8StringEncoding];
} else {
    [self sendStringToManagementSocket:[NSString stringWithFormat:@"password \"Auth\" \"%@\"\r\n", escaped(myPassword)] encoding:NSUTF8StringEncoding];
}
```

gbdlin:
It's much better to leave separate dialog for oauth token for cleaner connection flow, when username and password are saved in keychain. For flow when credentials are not saved in keychain this is not really a problem, because OpenVPN client displays it in new window anyways, unless you've been disconnected and it tries to reconnect.

timja-kainos:
I'm happy to test this out if someone can provide a signed built version of it.
I compiled the previous PR but had issues with mac blocking the running of it because it wasn't signed by a trusted certificate

jkbullard:
@faucct - **Thanks!** This looks great except for the following:

 1. Please use Cocoa's CommonCrypto  instead of MF_Base64Additions. There's a [stackoverflow thread](https://stackoverflow.com/questions/11386876/how-to-encode-and-decode-files-as-base64-in-cocoa-objective-c) with a simple implementation in an answer by denis2342. I'd implement it as-is in Tunnelblick's existing sharedRoutines.h/m (in the "Common" group), which would mean that the `base64String` use would need to change to use `base64enc()`. I'd also add NSLog() statements if `transform` was nil or if `SecTransformSetAttribute()` returns nil. (Tunnelblick almost always "logs and recovers" instead of doing an assert.) And if `base64helper()` returns nil `base64enc` should return an empty string.
 
 2. Please implement the `echoResponse` parameter: Make the input field in -getResponseFromChallenge:… either an NSSecureTextField or an NSTextField, depending on the value of `echoResponse`, and include the assignment and autorelease of `input` in the declaration line. Something like:
```
id input = (  echoResponse
	    ? [[[NSTextField       alloc] initWithFrame:NSMakeRect(0, 0, 200, 24)] autorelease]
	    : [[[NSSecureTextField alloc] initWithFrame:NSMakeRect(0, 0, 200, 24)] autorelease]);
```

 3. Since we try to avoid asserts, please change the line
```
    NSAssert1(NO, @"Invalid input dialog button %ld", (long)button);
```
to:
```
    NSLog(@"Invalid input dialog button %ld", (long)button);
```

I could merge this PR and make the changes myself afterwards, or you can update this PR and I'll merge it then. I'd prefer the former, but it's your choice, @faucct.

@timja-kainos - Once it's done I will create a signed release for testing.

faucct:
@jkbullard, I am okay with leaving the rest to you :)

jkbullard:
@faucct - Thanks again!

@timja-kainos - I will post here when a signed test version is available

timja-kainos:
@jkbullard Any chance of getting a test version? Thanks

PR #529, opened 2019-04-12
trust and upgrade xib to the modern format

m5o:
Hi @jkbullard 

I've just opened all the `xib` files once and saved them immediately without any layout change to the content. My intention is to separate the layout refinements from this (mandatory) update task.

* trust and upgrade `xib` to the modern format
* opend via Xcode `Version 10.2 (10E125)`
* **no** layout changes were made to the content

Screenshot 2019-04-12 at 13 16 54

🤔 Just looking at the LOC that will be added / removed `+2,529 −25,741` I hope everything is fine with this new 'modern format'. 🤷‍♂️


m5o:
@jkbullard any feedback from your standpoint? Why’re some of the project files outdated and never touched with a newer Xcode version lately, is there a reason?

jkbullard:
Project files are not updated because stable releases are built with Xcode 7.3.1 on macOS 10.11.6. (We did update the .xcodeproj file partially for Xcode 10.1, but the changes do not interfere with using it with Xcode 7.3.1.)

We build stable releases with Xcode 7.3.1 because newer versions of Xcode build binaries that get segmentation errors (SIGSEGV). We've never figured out what causes this, but eventually Apple updates Xcode to a version that generates code that works properly and we update to that version.

Most recently, for example, recent beta builds by Xcode 10.1 on macOS 10.13 were reported to fail with a SIGSEGV by about 0.04% of users (4 per 10,000). After several iterations of betas, we reverted and built the latest beta (3.7.9beta09) with Xcode 7.3.1.

Issue #528, opened 2019-04-11
apply Apple’s Human Interface Guidelines to Preferences


m5o:
Hey Tunnelblick Team 👋 

**Describe the “bug"**
I use Tunnelblick on a regular basis and have to open the settings from time to time. 
Coming from a design background my eyes cannot unsee some of the unaligned labels, backgrounds, and checkboxes in the preference window. Sorry 🙈

Before, I would like to ask if such contributions are welcome? 
I don’t know any ObjC and would mainly tweak the `*.xib` files with Xcode.

**Expected behavior**
Polish and refine spacing and position of Tunnelblick preferences.
📝 https://developer.apple.com/design/human-interface-guidelines/macos/app-architecture/preferences/

No changes to behavior and wording, only polishing the arrangement.

🤓✌️

jkbullard:
Such contributions would be very welcome.

However, please understand that the window contents are arranged to enable us to replace the English strings with strings in any of the 40 languages to which Tunnelblick is translated.

Instead of using the standard scheme of creating a .xib for each language, which means that translators must work with .xibs, Tunnelblick has a single .xib for each window [1]. When the window is loaded, Tunnelblick dynamically replaces each English string in the .xib with the corresponding string from the language currently being used. This means our translators only need to work with strings (using [Crowdin](https://crowdin.com)) – they don't need to know how to edit .xibs.

We leave a lot of space for strings because a string in English may be translated to a longer string in some other language. (And yes, this makes our windows much larger than they need to be in ideographic languages such as Chinese, whose translated strings are much shorter than the English strings they are based on.)

[1] Technically, we have two .xibs per window, one for left-to-right languages such as English, and one for right-to-left languages such as Arabic. The .xibs are more-or-less mirror-images of each other. And the .xibs themselves do not contain the English strings; they just contain placeholder strings that are used for layout. When Tunnelblick loads a window, it replaces these placeholder strings with the appropriate string from the current language, which might be English or might be some other language.

m5o:
Would it possible to use [Previewing Localizations in Interface Builder](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/TestingYourInternationalApp/TestingYourInternationalApp.html) to test these 40 available localizations?

<img width="562" alt="Screenshot 2019-04-12 at 13 39 31" src="https://user-images.githubusercontent.com/26371/56034727-6ff65380-5d28-11e9-8ddc-05749b050c9c.png">

[...]

PR #537, opened 2019-05-21
reset preferences window animation

### LINKED ISSUE #522: Text from two settings screens is displayed
**Describe the bug**
When switching between settings windows, content from previous window is still displayed when new one is shown.

**To Reproduce**
1. Open settings
2. Click on tabs
3. See error like screenshot

**Expected behavior**
No overlay

**Screenshots**
image

**Log**
I don't include log, since it is not related to VPN connection. If information is needed, you can take it here https://github.com/Tunnelblick/Tunnelblick/issues/521

m5o:
* reset preferences window animation time
* fix overlay content issues when navigation between
* 📕 Apple HIG
  * https://developer.apple.com/design/human-interface-guidelines/macos/app-architecture/preferences/
  * https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/toolbars/
    > Clicking a toolbar item results in an immediate action—such as opening a new window, switching to a different view, [...]—so it doesn’t make sense to imply that there is also a change in state.
* fixes #522

🔗 #528

### DIFF
--- a/tunnelblick/DBPrefsWindowController.m
+++ b/tunnelblick/DBPrefsWindowController.m
@@ -421,8 +421,8 @@ static DBPrefsWindowController *_sharedPrefsWindowController = nil;
     if ([self shiftSlowsAnimation] && [[[self window] currentEvent] modifierFlags] & NSShiftKeyMask)
 		[viewAnimation setDuration:0.00];
     else
-	
 		[viewAnimation setDuration:0.00];
+	
 	NSDictionary *fadeOutDictionary = [NSDictionary dictionaryWithObjectsAndKeys:
                                        oldView, NSViewAnimationTargetKey,
                                        NSViewAnimationFadeOutEffect, NSViewAnimationEffectKey,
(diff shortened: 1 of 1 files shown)


m5o:
@jkbullard I know, is a dirty hack resetting the animation duration. What is your opinion on this?

jkbullard:
I agree, resetting the duration to zero is pretty hackish; one might just as well disable the code entirely. (Or even strip out the entire DBPrefsWindowController/MyPrefsWindowController classes, because as I recall the reason they were used was to do this animation.) So I'd prefer something that instead fixes the problem but preserves the automation; that is, that fades the old view out and then fades the new view in. (I'm assuming that the current code does both simultaneously and that's what's causing the problem.) That may be more trouble than it's worth but should be looked into.

I haven't seen this problem myself, although I've seen the screenshot in #522 and a movie of it happening, so it may depend on hardware speeds or something like that. (Although why Apple would make animation that depends on hardware speed I don't understand. Maybe it's due to a change in macOS's automation instead?)

m5o:
Thank you for this detailed explanation. 🙇‍♂️

I do not have the Obj-C skills to refactor all the `DBPrefsWindowController.m` code, so I have to step back on this. Although remember the guidelines suggest changing different pref views without showing a "change in state".

PR #540, opened 2019-05-24
remove fixed application name suffix from preferences

m5o:
* remove fixed application name suffix from preferences
  * eg. `Info - Tunnelblick` -> `Info`
* 📕Apple HIG
  * https://developer.apple.com/design/human-interface-guidelines/macos/app-architecture/preferences/
  > **Update the window's title to reflect the currently visible preference pane.**
For example, if your preferences window has a General preference pane, the window’s title should be General when that pane is active.

🔗 #528

### DIFF
--- a/tunnelblick/MyPrefsWindowController.m
+++ b/tunnelblick/MyPrefsWindowController.m
@@ -1501,7 +1501,7 @@ static BOOL firstTimeShowingWindow = TRUE;
         appName = [appName substringToIndex: [appName length] - 4];
     }
     
-    NSString * windowLabel = [NSString stringWithFormat: @"%@ - Tunnelblick", localizeNonLiteral(currentViewName, @"Window title")];
+    NSString * windowLabel = [NSString stringWithFormat: @"%@", localizeNonLiteral(currentViewName, @"Window title")];
 
     if (  [currentViewName isEqualToString: NSLocalizedString(@"Configurations", @"Window title")]  ) {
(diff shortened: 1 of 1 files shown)


jkbullard:
Thanks, but I don't think this should be done.

The reason "Tunnelblick" is added to the window title is that otherwise there is no way to know what program is showing the window. In an "application" you can find out what program is showing the window by clicking anywhere in the window; the name in the menu bar will show the application name. But for a program such as Tunnelblick, which is not technically an "application" (because it lives in the status bar) that behavior doesn't happen.

The reason "Tunnelblick" is added at the _end_ of the window title is twofold:
 1. Historical: years ago most applications did this;
 2. Because most users would know what program the window belongs to, and what they want to know more urgently is what pane they are looking at or what configuration's information is being shown, so that information comes first.

Also please note that:
 * This is not, strictly speaking, a "Preferences" window. It is Tunnelblick's main all purpose window.
 * Even Apple doesn't follow their guidelines for Preferences windows. See, for example, Finder's Preferences in macOS Mojave.

m5o:
Ok, fair and valid points. I can follow your explanations. Of curse, in the end, the HIG is just a "Guideline" and yes it seems also Apple itself isn't following these sometimes. Hopefully, in the long run, mac developers will follow the guides, refine historical decisions and contribute to a consistent and modern user interface. With this change, I've just tried to catch the low hanging fruits first. 

Thank you for the detailed feedback

PR #544, opened 2019-06-03
fix Warnings label alignment

m5o:
* fix Warnings label alignment

Screenshot 2019-06-03 at 11 29 23

| before  | after |
|---|---|
| Screenshot 2019-06-03 at 11 11 57  | Screen Shot 2019-06-03 at 11 18 54 AM |

🔗 #528


jkbullard:
Great! Thank you!

PR #551, opened 2019-06-06
Optional challenge secret store in keychain

hryamzik:
Following #430, secret can be posted in a challenge input alert, `secret` button appears when `alt` key is pressed. That's more secure then getting OTP from a script.


hryamzik:
cc @jkbullard

jkbullard:
@hryamzik - Thanks. I'm sorry and apologize for taking so long to respond to this.

I think I now understand more about TOTP and am willing to save the "secret" in the Keychain. Here are some questions and suggestions, but you probably shouldn't actually **do** anything until I've agreed to your approach.

(A) Can you please expand on what the "Secret" button does in the commit message?

If "challengeSecret" for the configuration*is stored in the Keychain,  I _think_ it

 1. Retrieves "challengeSecret" the Keychain;
 2. Generates a time-based OTP from it;
 3. Uses the OTP for the response to the OpenVPN server's challenge.

If the user's response is **not** stored in the Keychain,  I _think_ it

 1. Saves the user's response to the challenge as "challengeSecret" in the Keychain;
 2. Generates a time-based OTP from the response;
 3. Uses the OTP for the response to the OpenVPN server's challenge.

(B) I gather the AeroGear code was taken from somewhere (it's copyrighted by Google). Can you provide a link to the source?

(C) Taking the comment "//  Taken with small modifications" in AGStringEncoding.m as an example, can you describe the "small modifications", or were they already present in what you downloaded? My preference, and what Tunnelblick does for most third-party code, is to include the code as originally provided, and use .diff files to patch it. (See, for example, how we build the tun/tap kexts.) For code that is integrated into the Tunnelblick application itself, I would prefer that externally-generated code first be committed as it comes from the source, then make changes to it in a separate commit. That clarifies exactly what's changed from the original code. If the original code you downloaded was already modified, please consider tracking down the code before it was modified.

(D) I'd prefer more granular commits. For example
 1. Add AeroGear code (as is, citing the source).
 2. Patch AeroGear code.
 3. Add changes to implement saving challengeSecret (n AuthAgent).
 4. Add UI changes that use 1. - 3.

(E) Separate from what the button does, I don't like it being revealed only if the "Option" key is pressed. I strongly dislike such hidden affordances. Tunnelblick doesn't have any hidden affordances. I know that Apple has lots of them, and I hate theirs, too.

(F) (Related to (E)). If I understand what this code does, I would suggest a different UI: add an always-visible checkbox that says "Save in Keychain and use TOTP", with a longer explanation in the checkbox's popup help. (Or a similarly-labelled button.) This would be much more in keeping with the way Tunnelblick does things, would avoid the hidden affordance, and would show the user a better explanation of what it does. (I'd probably want to also add an "eye" icon to toggle making the user's response visible like the username/password dialog box has, but that is a different topic and should be a different PR.)

jkbullard:
The Aerogear code in this PR is almost identical to the most recent commit ([76a4b623d](76a4b623d458ca8644366a4d611eb74ba481ca7d)) from https://github.com/aerogear/aerogear-ios-otp. That provides a git history. There are some changes to a few files, adding code to deal with reference counting, and two added files, MF_Base64Additions.h and .m.

That respository has "been deprecated and is not being maintained". I'm not sure why, but assume it is because the code has been rolled into AeroGear's "Identity Management" product. It's not a huge amount of code and seems reasonably straightforward, so I'm willing to add it to Tunnelblick assuming it passes my more thorough examination.

How to do the UI for this is still to be decided. Separate PRs for the different parts could get things moving before that's resolved:
 1. PR for adding Aerogear code "as-is". (Note that Tunnelblick's practice is to put .h and .m files directly into the "tunnelblick" folder, not into a subfolder as this PR does.)
 2. (Probably followed by patches to clarify the source of the code, citing the above GitHub repository so the history can be examined.)
 3. PR for adding the ability to store/retrieve a challenge secret in the Keychain.
 4. PR for the the UI.

hryamzik:
Sorry for late response, just got back from vacation.

> (A) Can you please expand on what the "Secret" button does in the commit message?

> If "challengeSecret" for the configuration*is stored in the Keychain, I think it

> Retrieves "challengeSecret" the Keychain;
> Generates a time-based OTP from it;
> Uses the OTP for the response to the OpenVPN server's challenge.

Correct. In this case OTP dialog is not displayed.

> If the user's response is not stored in the Keychain, I think it

> Saves the user's response to the challenge as "challengeSecret" in the Keychain;
> Generates a time-based OTP from the response;
> Uses the OTP for the response to the OpenVPN server's challenge.

Right. If you press `alt` and hit `secret` button. Otherwise it's a regular OTP input dialog.

> (B) I gather the AeroGear code was taken from somewhere (it's copyrighted by Google). Can you provide a link to the source?

I've actually pulled it from a previous PR, here's the link: https://github.com/Tunnelblick/Tunnelblick/pull/430#issuecomment-361086180

> (C) Taking the comment "// Taken with small modifications" in AGStringEncoding.m as an example, can you describe the "small modifications", or were they already present in what you downloaded? My preference, and what Tunnelblick does for most third-party code, is to include the code as originally provided, and use .diff files to patch it. (See, for example, how we build the tun/tap kexts.) For code that is integrated into the Tunnelblick application itself, I would prefer that externally-generated code first be committed as it comes from the source, then make changes to it in a separate commit. That clarifies exactly what's changed from the original code. If the original code you downloaded was already modified, please consider tracking down the code before it was modified.

That comment is part of the [code I've imported](https://github.com/aerogear/aerogear-ios-otp/blob/master/AeroGearOTP/Classes/AGStringEncoding.m#L2)

> (D) I'd prefer more granular commits. For example

> Add AeroGear code (as is, citing the source).
> Patch AeroGear code.
> Add changes to implement saving challengeSecret (n AuthAgent).
> Add UI changes that use 1. - 3.

So in this PR:
- Implemented challenge secret store in keychain 0972bf8
- Added check for missing secret in keychain 4f5d75b
- Fixed reference counts d952622

Do you want to just split the first commit?

> (E) Separate from what the button does, I don't like it being revealed only if the "Option" key is pressed. I strongly dislike such hidden affordances. Tunnelblick doesn't have any hidden affordances. I know that Apple has lots of them, and I hate theirs, too.

I think it should be more or less hidden from regular users. If you search how to store the secret – you find it (there's another less secure way to do it BTW), otherwise input codes. That may also save TB from being banned by security teams.

hryamzik:
cc @jkbullard

PR #561, opened 2019-07-05
Fix a translation error

ethicnology:
network = réseau =/= réseaiu

Sorry crowdin procedure is overkill just for a letter

### DIFF
--- a/tunnelblick/FR.lproj/Localizable.strings
+++ b/tunnelblick/FR.lproj/Localizable.strings
@@ -192,7 +192,7 @@
 
 /* Button
    Menu item */
-"Re-enable Network Access" = "Réactiver l'accès réseaiu";
+"Re-enable Network Access" = "Réactiver l'accès réseau";
 
 /* Button */
 "Reinstall" = "Réinstaller";


jkbullard:
@ethicnology - Thanks, however, I am rejecting this PR for two reasons:

 1. As you apparently know, we do translations on Crowdin and do not accept PRs for them. That's because we allow only a one-way flow of translations, from Crowdin to the source code. That lets us use the knowledge of our translators, who vote on translations, instead of having to rely on my knowledge of a language to evaluate whether or not a proposed translation is better. Since I don't know any French beyond a few words and phrases, I can't evaluate your proposed translation.

 2. The word "network" is used in more than 20 other strings, so changing it in one string will mean it is used inconsistently. If changed in this one string, it probably should be changed in most or all of the other strings, too.

If you would like to join Crowdin and do the work there, please email developers@tunnelblick.net and include the information described in [Signing Up to Translate for Tunnelblick](https://tunnelblick.net/cUsingCrowdin.html#signing-up-to-translate-for-tunnelblick). Please include the fact that you are "ethicnology" on GitHub.

If you do sign up, I would suggest you propose this one change and get the reaction of the other French translators before proceeding to change all of the strings that contain the word "network". (If you can contact all the other French translators through Crowdin's "Conversations" feature, you might want to do that instead.)

If you don't sign up in the next week, I will pass on your suggestion to our French translators and they will make the necessary changes if they agree with you.

Thanks again, though, no matter how it turns out.

Issue #407, opened 2017-10-20
Add authentication from users through Touch ID.


ericpardee:
Haven't looked how to implement in Tunnelblick,
 but it seems that The Local Authentication framework provides facilities to request that a user may authenticate using only Touch ID, i.e.
 during installation. (https://developer.apple.com/documentation/localauthentication)

tmthecoder:
@jkbullard Hello, I'm looking into adding this feature and submitting a PR in the future. Do you mind giving me a bit of a rundown through different instances of User Authentication? I just want to make sure I don't miss anything when covering my bases.

jkbullard:
Is the idea to:
 1. Have Tunnelblick optionally require a fingerprint to connect a VPN?
or
 2. Have Tunnelblick accept a fingerprint for authorization for installing itself (and configurations, etc.)?

The first one seems doable, but it wouldn't interact with the existing authentication, which, being managed by OpenVPN, is username/password based and/or certificate based. It would just be an additional step (optionally) required while starting a connection. (Would it be required for restarts, too?)

However, I don't understand why the second one would be necessary or how Tunnelblick could do it. For installations, Tunnelblick isn't asking the user to identify himself/herself. Instead, Tunnelblick asks macOS to authorize it to have "root" privileges for certain operations (e.g., installing Tunnelblick itself, installing configurations, making a change to a configuration, etc.). macOS then puts up a prompt asking for the username/password of a computer administrator. if they are correct, macOS then gives Tunnelblick an authorization token that Tunnelblick uses when it needs to perform a privileged operation. Because the authorization is done by macOS, I assume that on a Touch ID mac, you can use a computer administrator's fingerprint instead of his/her username/password. Is that not the case?

tmthecoder:
@jkbullard My initial start was to do an optional one as you stated in the beginning, and that I have a plan laid out(going to implement soon) as it does not interact with any existing authentication. The second one was what I thought might be possible, as we may be able to use touchID for whatever operations Tunnelblick needs to run as admin, which I have to do more research on, but if it runs through keychain you can use touchID to do it(some things touchID does not allow over password entry to my belief, I have yet to see what specifically)

jkbullard:
The second one does not have anything to do with Keychain. To get the authorization, Tunnelblick uses a system call, AuthorizationCopyRights(), in Tunnelblick's SystemAuth class's checkAuthorizationRef:prompt:interactionAllowed:reactivationAllowed: method.

[...]

PR #585, opened 2019-11-17
Optionally Require Authentication to Connect (Using TouchID if available)

tmthecoder:
Hello, I added the functionality to use touchID in order to authenticate and start a VPN connection. This is extra functionality added through the LocalAuthorization framework and will not cause any backward comparability problems. The feature's enablement switch is found within the Connecting & Disconnecting section and is a simple checkbox, storing the preference in the same way as all other items. Upon enabling this feature(per-vpn) it will now ask you to authenticate with touchID when connecting to the VPN that the checkbox was selected on. I have broken it into 10 small commits, of which the first 6 or 7 are mainly streamlined changes and the last couple are me fixing some errors/bugs that I overlooked. I hope this can be a nice feature used, and please let me know if I need to change anything 

DISCLAIMER: this is my first PR, so if I did something wrong, please let me know as it's my first venture into contributing to open source and I would like to get better at it 

Thank you!


jkbullard:
@tmthecoder - Thank you very much. This is a great first PR. (Especially considering how quickly you created it!) I appreciate the small commits, which make it easy to examine.

Here are my **very** detailed comments. Please keep in mind that I am **very** picky; please don't be offended or embarrassed or anything.

Note: This is all based on testing on my 2012 MBP with Mojave 10.14.6. It does not have a touchbar. I don't have access to a touchbar Mac, so I can't test that part but assume you did.

First, a question: After checking the box, when I connect a window pops up asking me to authenticate by entering a password, which seems reasonable. **Does the same window pop up on a TouchID-equipped Mac, and the user can either enter the password or provide their fingerprint?** I'll assume that's the case. And I'll assume that in general when macOS pops up such a window it will accept either entry of the password or the fingerprint. (Otherwise, what happens if your finger is amputated?)

Now, my comments:

**1. Serious**: If I cancel that window, then all subsequent attempts to connect are just ignored and nothing appears in the system.log. (Which is odd, because it looks like cancellations should be being logged.) Even if I un-check the box, or un-check it and re-check it, attempts to connect are ignored. I have to quit and relaunch TB to get the ability to connect; then an attempt to connect asks me to authenticate (if the box is checked) or just connects (if the box is not checked). So maybe the "cancelled" state is being preserved from one connect attempt to the next, but cleared when Tunnelblick relaunches. It is _possible_ that this is a flaw in Tunnelblick outside of the code in your PR, but please look in your code first.

**2. Serious*: Assuming the purpose of this is to force the logged-in user to authenticate before connecting the VPN, it doesn't do that because anyone could un-check the box and connect without authenticating. It protects against someone clicking "Connect" by mistake, but I don't think that's very useful.

Making the checkbox protected so that only a computer administrator can check/uncheck it is probably the best solution. That can be done by making authenticateOnConnectWasClicked: work the way that generalAdminApprovalForKeyAndCertificateChangesCheckboxWasClicked: does. (It creates a new thread to authorize the change and undoes the change to the checkbox; the thread changes the checkbox and the protected preferences only if the user authorizes the change.)

Another solution would be to require the logged-in user to authorize changing the checkbox, but that doesn't allow corporate admins to require the authorization (which I think would be good functionality to add). (They sometimes don't allow their users to be administrators on their computers, and allowing an ordinary user to do it defeats that.)

**3. Maybe serious or nothing**: Is the authentication being done at the right point in the connection process? That is, is anything being done before that needs to be undone if the auth fails or is cancelled? I'm not sure about that and it will require a lot of staring at the code, so it would help to have (at least) two of us doing that.

**4.**: If I try to connect with the box checked, a window pops up saying "Tunnelblick is trying to Authenticate using TouchID". I assume the "Authenticate using TouchID" string comes from your VPNConnection.m:

`NSString * promptMessage = NSLocalizedString(@"Authenticate with TouchID", @"TouchID Prompt");`

I think it should just be "authenticate" (not capitalized and not mention TouchID). If you need the "using TouchID" to make it clear to TouchID users that they can use TouchID, the string should be determined at run time depending on whether or not TouchID is available.

**5.**: The checkbox title should either be something more like "Authenticate before connecting" (less wordy), or maybe "Authenticate before connecting using TouchID" if you absolutely _must_ mention TouchID. If the authentication window always allows Touch ID to be used, then it probably needn't be mentioned. So maybe the string depends on TouchID being available.

**6.**: The second commit removes the "Load Tun automatically$" title and the "state=on" for it from the .xib. I assume this was an inadvertent change but it should be reversed.

**7.**: I think there's a memory leak in your getAuthentication method: "LAContext * context" is never released. I'm not sure (because of the blocks and dispatch), but maybe you can just add autorelease to it, so it is

`LAContext * context = [[[LAContext alloc] init] autorelease];`

Otherwise you'd have to release it manually at some point.

**8.**: I think the last commit, "Changed the localized string to reflect the necessity of touchid…" is based on a wrong idea, that non-touchID devices (versions of macOS?) can't do the authorization. As described above, my MBP doesn't have TouchID but can do the authorization.

I think that on a system that doesn't support LAContext (lower than Yosemite, apparently) the checkbox should be forced un-checked and disabled, and the infoTitle should include a paragraph like "Disabled if your version of macOS does not support it.".

To decide if LAContext is supported, you can either use Tunnelblick's runningOnYosemiteOrNewer() or a similar function in helper.m, or test it is supported directly. I don't remember how to test directly. I think you see if you can create a class object from its name (as a string) but am not sure.

**8.**: The shouldAuthenticateOnConnect method should return no if the system doesn't support LAContext. That handles the situation of somebody importing preferences that have the checkbox checked onto a system that doesn't support LAContext.

**9.**: You couldn't know this, but it isn't necessary to modify English.lproj. Our release process includes an automated procedure that modifies that file by parsing the project's source code to extract all the strings that need localization. But doing the changes to English.lproj in the PR doesn't hurt, and it makes the PR more complete, so there is no need to do anything about it.
 
Whew! Thanks again!

tmthecoder:
@jkbullard Thanks for the help! I'm working on adding some of the changes, but regarding #7, I believe I already have autorelease in the code. It shows up in my GitHub repo, as well as in the commit for the PR. Does it not show up for you?

jkbullard:
You're right. I must have been looking at one of the early commits or something. It does have an autorelease. Sorry!

tmthecoder:
Question regarding #2, Should I move the definition of the `authenticateOnConnectWasClicked` method to `MyPreferencesWindowController`? Or use the code from the `generalAdminApprovalForKeyAndCertificateChangesCheckboxWasClicked` in the same `SettingsSheetWindowController`?

jkbullard:
@tm - Don't move to `MyPreferencesWindowController`. That's for everything in the "VPN Details" window. Everything for the Advanced Settings window goes in `SettingsSheetWindowController`. Sorry I forgot to mention that.

(I haven't looked at your recent commits yet. I hope to do that later tonight or tomorrow.)

tmthecoder:
When writing to a preference securely, do I need to make a new forced path like this one: `L_AS_T_PRIMARY_FORCED_PREFERENCES_PATH`?

jkbullard:
No. The forced preferences are a .plist which may not exist and may have zero or more preferences. All of the forced preferences (one, and two when your PR is done) are in the same .plist.

The code reads in the .plist, adds or replaces the value of the preference, then asks for authorization. If it gets the authorization, it writes out the .plist (and sets/clears the check in the checkbox. At least that's what it was supposed to do!

I think there is the possibility of a weird race condition now that there are two preferences that are protected: that the .plist is read in, and, while waiting for authorization on that, the other checkbox is clicked and the .plist is read in again. Then, depending on which authorization is given first, one of the changes will be lost. Or maybe macOS will allow only one authorization dialog to be shown at a time? In any case, I think this is such an odd, easily avoided race condition that we can ignore it for now. I just wanted to document that it does exist.

tmthecoder:
I currently used the same code, and committed it, but it is giving me a false write no matter what with the checkbox temporarily going checked, but blank again. Upon checking the file it is also written as false. In order to read them in, or write to them, do I need to set a special parameter or anything?

jkbullard:
Ah. It isn't as simple as I thought. You not only need to have something similar to `generalAdminApprovalForKeyAndCertificateChangesCheckboxWasClicked`, but it must detach  a new thread like `generalAdminApprovalForKeyAndCertificateChangesThread`, and all of the things it invokes, and a new method like `setupUpdatesAdminApprovalForKeyAndCertificateChangesCheckbox`.

So I can see that it is very complicated. It is probably better to have the existing `generalAdminApprovalForKeyAndCertificateChangesCheckboxWasClicked` and your new `authenticateOnConnectWasClicked ` both invoke more generic methods.

If you want to do that (essentially refactoring `generalAdminApprovalForKeyAndCertificateChangesCheckboxWasClicked` and everything it invokes to be more general and then invoking those more general methods), that would be great; but you'll have to trace what's going on, which is complicated.

So if you'd like me to do that and commit it to master, you could then pull that in and start using the more new generic method(s). Let me know if you want me to do that; I think I could do it by Thursday.

This may be taking more time than you (or I) thought initially, so thanks for persevering. I think this will be a great feature.

tmthecoder:
I have no problem moving (or at least trying to) move them into a more generalized pattern, but what file would you want me to move them to?

tmthecoder:
And I have no problem waiting, I'm excited to be helping out in any way, especially because I've used Tunnelblick for so long and glad I decided to start contributing

jkbullard:
Good question! UIHelper, I guess. It has mostly display-related methods (that are used as if they were subroutines), but these protected preferences are related to the UI.

UIHelper has only class methods currently (no instance methods), so that would be the best way to go if you can, and have things passed around in the dictionary (maybe an outer dictionary with parameters which contains the existing dictionary that contains the .plist info). But if you have a better idea, go with it!

tmthecoder:
So here's my thoughts on how this would work: I would make a public writer method within the UIHelper class, then mimic the same as the normal writer would do, with the parameters of a key (other parameters I'm missing?) And make the other thread creator and other selector methods private. The ones requiring UI changes would be written within their respective files, but main Auth would be handled from the UIHelper class. Basically to the point where we get a finished result and set the checkbox to said result from it's respective UI file

tmthecoder:
Please correct me if I'm wrong in any instance or doing something that will cause any security or other problems

jkbullard:
I think that sounds great.

jkbullard:
So the preference keys is one parameter, another would be the checkbox, so it could be checked or cleared? Not sure you can easily put the checkbox object object in a dictionary, so maybe just the key, and then the method looks at the key and then chooses which checkbox to check/clear. (That's a little nasty, but might be better than somehow encoding the checkbox.)

tmthecoder:
Okay. Here's a bit of an update: I figured out what was wrong with my localized method (the city of the `generalAdminApprovalForKeyAndCertificateChangesCheckboxWasClicked` method). That ended up being the inverse state that the general one uses which mine doesn't. That works, but when trying to use the unified method, I get the error: `target does not implement selector`. using the same code from the selector function that was put in the error (UIHelper::secureAuthThread) works within the local file, so it may be a problem with how I structured things. I'll work on more of that, but would you like to submit that as a separate PR? Currently the touchID/auth on connect features should be ready. Regardless, I can keep working (or try my best) on the unification of the two secure writing methods

tmthecoder:
The selector problem may have to do with the fact it is not declared as a object, but follows the static context of the other methods, but I'm not absolutely sure.

jkbullard:
1. A separate PR for having the checkbox be protected would be fine.
 2. I think you can pass a selector as the string that names it, and then convert that string back to a selector.

tmthecoder:
For now, the box is protected, it's just not in the general method for both. They both use the same method in different files. Is it possible to keep it like that for now and then work on a separate PR to merge the two secure write functions?

jkbullard:
OK, I misunderstood (or lost track). Sure, that's fine, as long as it works. It makes sense to have it working and then refactor the nearly duplicate methods into generalized methods.

tmthecoder:
Yeah I get it this turned out to be quite big, and I'm glad to be a part and help. I'm assuming we can turn the generalized methods into a separate PR? As of now, I believe all 9 points in the initial review have been fixed from the latest commit

jkbullard:
Sure, a separate PR for the generalized methods seems quite appropriate. I hope to look over the new commits and test it tomorrow or Wednesday. Thanks again!

tmthecoder:
Yeah no problem! I'm glad to help after using Tunnelblick for a while, but my knowledge of objective-c was very low until recently

tmthecoder:
Also, the commits after we spoke about generalizing the methods are still there code-wise, but not being used. The code I left in UIHelper(the 5 commits at the end excluding the last) may be of help in generalizing the methods, but I can remove them if needed

jkbullard:
We should do the generalization at some point, even if we don't do the "biometric but not password".

If it is easy to remove them from the PR and keep them for a later, separate PR, that would be great. Maybe just make a duplicate copy of the source code with the commits (i.e., copies of the files as modified with the commits), then remove the commits. I assume there's an easy way to do that with git commands, especially since they are the last few commits. Then in a separate PR, use the duplicate files and do commits with them.

But if it's not relatively easy, we can just leave them in the PR since they aren't being used.

tmthecoder:
I was definitely planning on doing (or helping with) the generalization at some point, and my thighs exactly were that the PR would be cleaner and more topic oriented if I temporarily removed the generalization code. I'll keep a copy of the files with generalization on my machine and commit them if needed in a later PR

jkbullard:
Great!

tmthecoder:
Okay, the code should be all set. I just removed the methods not necessary to touchID at this moment, but there is a backup on my MacBook. Hope everything's all good and as always let me know if anything needs fixing/improvement!

jkbullard:
Thanks, @tmthecoder. This is looking better and better!

Here is my second round of comments on this PR. It's based on staring at the code, without testing it. (And it's based on the code after 117316e1d2204364fb38e7eb73ba98a25606ac36).

Again, please don't be discouraged or embarrassed; I am very picky! Most of this is cosmetic or housekeeping; I only found one bug.

1. Please remove the following two lines from UIHelper.m (I think they were put in to prepare for the generalized methods for protected preferences.):

```
#import "SystemAuth.h"
#import "NSFileManager+TB.h"
```
2. Similarly, please remove the following single line from SettingsSheetWindowController:

```
//    [UIHelper secureOnClickWriter:newState key:[configurationName stringByAppendingFormat:@"-authenticateOnConnect"] inverted:NO];
```
3. SettingSheetWindowController.m line 1325 should be removed (I assume it was for debugging):

```
NSLog(@"%d", (int)status);
```

4. Please move the declarations of `shouldAuthenticateOnConnect` and `authenticateUser` from VPNConnection.h to VPNConnection.m, so they are private.

5. In the `authenticateUser` method of VPNConnection.m, please add language to the `NSLog()` calls to make it clear what succeeded or failed, and don't use all caps. For example, "User successfully authenticated" and "User failed to authenticate". It isn't necessary to add the name of the config that it applies to, but it might be nice and, if you do, you should use [self displayName] to get it.

6. If `[context canEvaluatePolicy:LAPolicyDeviceOwnerAuthentication error:&authError]` returns NO, the main thread will be stuck waiting for the semaphore to fire, but it will never fire. I think the solution is to move the `dispatch_semaphore_wait()` call to be inside the `if`.

7. If `[context canEvaluatePolicy:LAPolicyDeviceOwnerAuthentication error:&authError]` returns NO, the error should be logged, including all the error info provided by the NSError.

8. For clarity, and consistency with the rest of Tunnelblick, please do a global search/replace of `authenticateOnConnect` with `authenticateOnConnectCheckbox`.
    (Then do one to change `setupUpdatesAuthenticateOnConnectCheckboxCheckbox` back to `setupUpdatesAuthenticateOnConnectCheckbox` : )

9. Please change the `shouldAuthenticateOnConnect` method to a routine in helper.h/m named `localAuthenticationIsAvailable()`.That will make it available to SettingsSheetWindowController.m, allowing 10, below. You can simply have it return `runningOnYosemiteOrNewer()` for now. It may have to change because El Capitan wouldn't do this properly when I did a test. But maybe something else was wrong, so I'm not certain about this. 

10. Please make the checkbox consistent with other Tunnelblick checkboxes by disabling it when the feature is not available:
 * Set the enabled/disabled status of the checkbox in the `setupUpdatesAuthenticateOnConnectCheckbox` method and remove the two lines that enable it in the `finishAuthenticating:` method.
 * In the `enableEverything` method, invoke `setupUpdatesAuthenticateOnConnectCheckbox` so the checkbox is enabled only if appropriate.
 * I have learned the hard way to be particularly careful about any strings that users see, because several times our translators have pointed out errors I have made or suggested changes to improve clarity. (It's particularly embarrassing since I am a native English speaker and most of the translators are not!) Changing the English after it has gone out to the translators is particularly painful because it can mean a lot of translators have to re-translate a string to fix the English, even if their translations are correct.

    So, please change the infoTitle text for the checkbox () to the following:

```
@"<p><strong>When checked</strong>, you will be required to authenticate yourself before connecting. You can authenticate yourself by using your password or, if available, TouchID or FaceID.</p>\n"
@"<p><strong>When not checked</strong>, no authentication will be required before connecting.</p>\n"
@"<p><strong>This checkbox is disabled</strong> if you are using a version of macOS that does not support it.</p>"
```

I lower-cased "you", use "required" instead of "asked", clarified the language (you can connect without using a "Connect" button), added the "disabled" paragraph, added "FaceID" for future-proofing in case it becomes available on Macs, and made the language more vague about when ti is disabled to future-proof it against discovering that Yosemite doesn't fully implement it. (A quick test with El Capitan didn't work for me, so this is a real possibility.) Because this text appears in its own window, it isn't necessary to make it as concise as possible.

Again thanks for working on this!

tmthecoder:
All the code changes are relatively minor, and it should fix everything that was just asked. Sorry about the comments, I completely forgot to remove those and thanks for the help in fixing my code!

tmthecoder:
Okay. Now it should be all set to go. I forgot a semicolon, so it threw some things off but it should be fine now. Let me know if anything else is needed

jkbullard:
Thanks, @tmthecoder. Getting close!

1. I apologize for missing this the first time around, but the following line in UIHelper.m is for refactoring to use the generalized method and should be removed until that PR is done:
```
extern NSFileManager  * gFileMgr;
```

2. In 27a3c0f9, you moved both the semaphore wait and the semaphore release into the `if` block. You should only have moved the wait, because the semaphore is initialized outside of the block and failing to release it is not a good idea.

    Even better would be to leave the release in the block and move the initialization of the semaphore inside the block. And while you're at it, move the initialization of `promptMessage` inside the `if` block, too, since it isn't used outside that block.

3. Please undo the inadvertent change to the .xcodeproj in 1f93c726. 

4. In 86e32684510e6b2e5d05ab2b13e78e4f50afe6be, you mis-copy/pasted. The full statement should be:
```
    [authenticateOnConnectCheckbox setTitle: NSLocalizedString(@"Authenticate before connecting", @"Checkbox name")
                                  infoTitle: attributedStringFromHTML(NSLocalizedString(@"<p><strong>When checked</strong>, you will be required to authenticate yourself before connecting. You can authenticate yourself by using your password or, if available, TouchID or FaceID.</p>\n"
                                                                                        @"<p><strong>When not checked</strong>, no authentication will be required before connecting.</p>\n"
                                                                                        @"<p><strong>This checkbox is disabled</strong> if you are using a version of macOS that does not support it.</p>",
                                                                                        @"HTML info for the 'Authenticate before connecting' checkbox."))];
```

5. In `shouldAuthenticateOnConnect`, please change the call to `runningOnYosemiteOrNewer()` to call `localAuthenticationIsAvailable()` instead.

6. In `authenticateUser`, please change:
```
        NSLog(@"LAPolicy error: %@/", authError.description);
```
to:
```
        NSLog(@"LAContext canEvaluatePolicy error: %@", authError.description);
```

Again, thanks for working on this.

tmthecoder:
I just fixed those errors! I hope everything is in shape now!

jkbullard:
@tmthecoder - Great, thanks! Now I'll start testing it on various versions of macOS, which may take a couple of days.

tmthecoder:
Okay! Let me know if any support changes are needed (making the localAuthenticationIsAvailable to a higher version check or anything else. I've tested it on my MBP(15 inch 2018) and it works fine, let me know if I'll need to test anything touchID related

jkbullard:
@tmthecoder, what version of macOS did you test it on with TouchID?

I've run some tests on different versions of macOS in VMs on a 2012 MBP without TouchID.

It fails on macOS 10.10.5 (Yosemite) and 10.11.6 (El Capitan), see below, but on macOS 10.12.1 (Sierra) - 10.15.1 (Catalina) it works, showing a dialog with a textbox to enter a password. The dialog says:

    **"Tunnelblick" is trying to authenticate.**
    Enter password for the user "<LONG USERNAME>" to allow this.

Which makes a tiny bit of sense, because the first Mac with a Touch Bar was the Fall 2016 MBP which runs only Sierra and higher. It contradicts Apple's documentation (which says it should work on 10.11+), but that's nothing new.

**So I think you should change `localAuthenticationIsAvailable()` to use runningOnSierraOrLater().**

On macOS 10.10.5 (Yosemite), it fails with:
```
11/20/19 8:41:29.038 PM Tunnelblick[393]: Error Domain=com.apple.LocalAuthentication Code=-1001 "Unknown policy: '2'" UserInfo=0x60000027bf40 {NSLocalizedDescription=Unknown policy: '2'}
```

On macOS 10.11.6 (El Capitan), it fails with:
```
11/20/19 8:46:28.098 PM Tunnelblick[4641]: LAContext canEvaluatePolicy error: Error Domain=com.apple.LocalAuthentication Code=-1000 "Unexpected ACM requirement: 3" UserInfo={NSLocalizedDescription=Unexpected ACM requirement: 3}
```

tmthecoder:
I used Catalina. I will change it to return Sierra or later and push.

tmthecoder:
Just changed it!

jkbullard:
@tmthecoder - THANKS! I think you're done. I'll do final testing and merge this, probably tomorrow.

tmthecoder:
Okay! Awesome!

jkbullard:
Thanks again, @tmthecoder. This has been merged.

jkbullard:
Fixes #407.

PR #586, opened 2019-11-21
Add the checkbox for the authenticate on connect button to the RTL .xib or SettingsSheetController-RTL.xib

tmthecoder:
I added the same checkbox(authenticateOnConnectCheckbox) from the original LTR .xib file. It follows the same outlet, and works the same, to add support for authentication on a connection for those using Tunnelblick in RTL languages


jkbullard:
Looks great, thanks!

PR #608, opened 2020-03-10
Support SAML authentication with the Aviatrix Systems' VPN gateway

eliasaviatrix:
This code diffs enable Tunnelblick VPN client to support SAML authentication with the Aviatrix Systems' VPN gateway. 
 
The following describes how it works.
 
1. The VPN client scans the .ovpn file and retrieves the Aviatrix Systems authentication server URL.
 
2. The VPN client then connects to the server URL with one embedded web browser that is written in Apple webkit and The Aviatrix system then redirects the URL to a 3rd party SAML IDP (e.g. OKTA)
 
3. After the authentication is done, the client receives a message containing an email id and a token.
 
4. The VPN client then uses the e-mail id and token from the step 3 to connect to the Aviatrix System VPN gateway.

### DIFF
--- a/tunnelblick/ConfigurationManager.h
+++ b/tunnelblick/ConfigurationManager.h
@@ -88,7 +88,8 @@ typedef enum
 
 +(NSString *) parseConfigurationForConnection: (VPNConnection *) connection
 							  hasAuthUserPass: (BOOL *)          hasAuthUserPass
-						   authRetryParameter: (NSString **)	 authRetryParameter;
+                           authRetryParameter: (NSString **)	 authRetryParameter
+                                      samlUrl: (NSString **)     samlUrl;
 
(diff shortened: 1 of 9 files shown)


jkbullard:
I'm not inclined to merge this. I think a better approach would be to use a PAM plugin with OpenVPN directly. Is there some reason that can't be done?

eliasaviatrix:
I did not use the PAM. The hardest party to me is how to implement one embedded browser in PAM plugin to do the 3rd party web authentication (google, okta). The username and password is always changed for every login. Do you have a good reference github or document for this?
Do you have the plan to implement the 3rd party web authentication, e.g. Google auth, SAML auth... ? If you plan to make it with the OAuth ..., then I can wait or I can contribute some as well. 
We are the VPN vendors. We can see the need from the customers.

PR #660, opened 2020-12-28
make arm64 binaries everywhere

jim-lake:
arm64 arch for openvpn 2.4 and 2.5 and the tunnelblick process


jkbullard:
Thank you. I'll review it in more detail when I have a chance.

jkbullard:
Thanks again! I have merged this PR. Note that I will soon commit other changes that build all of Tunnelblick for arm64 and create Universal binaries.

PR #674, opened 2021-03-26
Support for security tokens added

macmoonshine:
- Added a checkbox, input fields and a switch for visibility to the login window
- The input fields are for entering security tokens (e.g. YubiKey) without changing the password.
- This allows the use of security tokens together with passwords from the keychain.
- The password for openvpn is simply created by appending the security token to the password.

### DIFF
--- a/tunnelblick/AuthAgent.m
+++ b/tunnelblick/AuthAgent.m
@@ -230,6 +230,7 @@ TBSYNTHESIZE_NONOBJECT_GET( BOOL,       showingPassphraseWindow)
     
     NSString * usernameLocal = nil;
     NSString * passwordLocal = nil;
+    NSString * securityTokenLocal = nil;
 
     if (  [self usernameIsInKeychain]  ) {
         usernameLocal = [usernameKeychain password];
@@ -254,9 +255,11 @@ TBSYNTHESIZE_NONOBJECT_GET( BOOL,       showingPassphraseWindow)
     }
 
     NSString * key = [[self displayName] stringByAppendingString: @"-alwaysShowLoginWindow"];
+    NSString * tokenKey = [[self displayName] stringByAppendingString: @"-useSecurityToken"];
     if (   (! passwordLocal)
         || (! usernameLocal)
-        || [gTbDefaults boolForKey: key]  ) {
+				|| [gTbDefaults boolForKey: key]
+        || [gTbDefaults boolForKey: tokenKey]  ) {
         
         // Ask for password and username
 
@@ -291,7 +294,8 @@ TBSYNTHESIZE_NONOBJECT_GET( BOOL,       showingPassphraseWindow)
 
         usernameLocal = [[loginScreen username] stringValue];
         passwordLocal = [[loginScreen password] stringValue];
-        
+        securityTokenLocal = [loginScreen useSecurityTokenChecked] ? [[loginScreen securityToken] stringValue] : @"";
+
         if (  ! usernameLocal  ) {
             NSLog(@"username is nil for Keychain '%@'", [usernameKeychain description]);
             usernameLocal = @"";
@@ -300,7 +304,10 @@ TBSYNTHESIZE_NONOBJECT_GET( BOOL,       showingPassphraseWindow)
             NSLog(@"password is nil for Keychain '%@'", [usernameKeychain description]);
             passwordLocal = @"";
         }
-        
+        if (  ! securityTokenLocal ) {
+					securityTokenLocal = @"";
+			  }
+
         if (   [loginScreen isSaveUsernameInKeychainChecked]  ) {
             
             if (   [loginScreen isSavePasswordInKeychainChecked]  ) {
@@ -345,7 +352,8 @@ TBSYNTHESIZE_NONOBJECT_GET( BOOL,       showingPassphraseWindow)
         [[loginScreen window] close];
     }
     
-    NSArray * array = [NSArray arrayWithObjects: usernameLocal, passwordLocal, nil];
+    NSString * passwordAndToken = [passwordLocal stringByAppendingString:securityTokenLocal];
+    NSArray * array = [NSArray arrayWithObjects: usernameLocal, passwordAndToken, nil];
     return array;
 }
 
--- a/tunnelblick/LoginWindow-RTL.xib
+++ b/tunnelblick/LoginWindow-RTL.xib
@@ -111,7 +111,7 @@ Gw
                                     <outlet property="nextKeyView" destination="86" id="104"/>
                                 </connections>
                             </button>
-                            <button fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="86">
+                            <button fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="86" customClass="TBButton">
                                 <rect key="frame" x="18" y="132" width="332" height="18"/>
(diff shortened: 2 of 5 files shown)


jkbullard:
Thanks for your PR. I will look at it in detail, but probably not for several days or even a week or two before I can do that, so please be patient.

I'll also have to think about whether I want to add more items to the already-complex username/password window.

Tunnelblick already has a mechanism which can be used to append a security token's output to the password before the password is sent to OpenVPN. You add (to the .tblk configuration) an executable (usually just a script) named "password-append.user.sh". Tunnelblick will run it as the user after getting the username and password (either from the login window or the Keychain). The executable (supplied by you) should get the token's current output (either by asking the user or by reading it directly from the device), echo it to stdout, and return with a status of 0 to indicate success (or return a non-zero status to indicate failure). Executables with other names can be used to prepend the token's output to the password, or to replace the password with the token's output. See [Tunnelblick VPN Configuration scripts](https://tunnelblick.net/cUsingScripts.html#tunnelblick-vpn-configuration-scripts) for details.

For users who save the password in the Keychain and have a machine-readable token, that mechanism provides a much better experience because no user interaction is required: Tunnelblick gets the username and password from the Keychain, gets the token's output from the executable (which gets it by reading the token hardware), and sends the username and the password (with the token's output appended to it) to OpenVPN.

But even though Tunnelblick already has this other mechanism, for some people your method is easier because it doesn't involve creating the executable, and/or is a nicer experience if they don't have a machine-readable token.

macmoonshine:
A script is probably not usable or very difficult to use because the security token requires user interaction. You have to press a button on the dongle so that it attaches the token to the password.
Anyway, I can understand your concerns about the dialogue. An alternative would be to move the switch for the field to the settings and only show the input field when it is switched on there.

jkbullard:
If you need to press a button on the security key to make it readable by the machine, a script could pop up a box instructing the user to push the button and have "Cancel" and "OK" buttons. The script could cancel, or read the value from the key and output it. That way the user wouldn't have to copy the value displayed on the key and type it in; they'd just click "OK".

But there are keys that are not machine readable, and your PR will work with them as long as the code needs to be appended to the password and not prepended to it.

jkbullard:
I have taken a close look at this. In general it looks great, so thanks again!

I've decided it is acceptable to add this to the login window itself, even if it increases clutter. It is much more self-contained that way. If that seems to be a problem, we can move it to the "Advanced" settings window later.

Here are my comments, based only on reading the code (that is, I haven't tested it):

First, some nit-picking cosmetic issues:
 * "securityTocken" should be "securityToken" in new lines 223, 228, and 229.
 * The empty line after new line 235 should be restored.
 * "password" should be "token" in the comments in new lines 339 and 345.
 
A slightly larger problem is that the .xib file was edited with a later version of Xcode than 7.3.1, which breaks compatibility with our standard build system. I can fix this after merging the PR easily 

The biggest problem (but easily fixed) is that you made changes only to LoginWindow.xib. Please make corresponding changes to LoginWindow-RTL.xib. That's the .xib which will be used for right-to-left-languages such as Arabic and Hebrew. It's basically a mirror image of the regular .xib. Again, you can use a later version of Xcode and I'll fix it to work with Xcode 7.3.1 after merging the PR if you don't have access to Xcode 7.3.1.

macmoonshine:
I have applied the suggested changes to the fork.

jkbullard:
Thanks, it works!

I played around with the checkbox and got myself confused at one point. So I suggest that, when the "Security token" checkbox is clicked, also do the following:

 * `visibleSecurityToken` should have it's string value set to @"". (This is what confused me. If you type a token value, make it visible, and then un-click "Security token", the value is there but it won't be used because the checkbox isn't checked.

 * `visibleSecurityToken` should be disabled or enabled depending on the checkbox value just like `securityToken`. Otherwise you can type in the field even though the checkbox is not checked.

 * `securityEyeButton` should be made not enabled and hidden, or enabled and not hidden, depending on the checkbox value. (That is, it shouldn't be visible or enabled when the checkbox is not checked.)

Some other points (not necessary for this PR, but feel free to add them if you want, in this or a separate PRs):

 * I'm not sure "Security token" is the best name for the checkbox.  It's longer than "Username" or "Password", and it's translation into one of our languages might be too large to fit into the limited space available. Maybe name it "Token" instead?

 * At some point it would be nice to add popup help for the checkbox (and the other checkboxes?). That would help users understand what will be done with the token. Maybe if/when that's done "Token" can be used as the label instead of "Security token". To do that, change the checkbox from an `NSButton` to a `TBButton` and set the title using `setTitle:infoTitle`. For an example, see the use of `routeAllTrafficThroughVpnCheckbox`.

 * It would also be nice to save the state of the checkbox in Tunnelblick's preferences when it's clicked, and set the state according to the preference in `redisplay`. That can be a future enhancement if you don't want to do it now. If you do, please suffix the configuration name with '-loginWindowSecurityTokenCheckboxIsChecked' as the key for the preference. The code for setting/getting the preference can mimic the similar code for `alwaysShowLoginWindowCheckbox`.

macmoonshine:
I have implemented the suggested changes, although I find it a bit unusual to hide the deactivated eye button as well. The value of the checkbox was already saved in the settings. I have only changed the name.

jkbullard:
Thanks!

I'm have merged this PR and will follow up with a commit that restores the ability to use Xcode 7.3.1, which can't handle the .xibs that were edited with a later version of Xcode.

I apologize for my suggestion that you also save and restore the state of the checkbox. **You already did that!** Without looking at the code, I just went by the behavior but must have just used "Cancel" instead of "OK", so the state wasn't saved. The name you had used for the preference was fine and much more in line with the other preference names. I'm sorry I didn't give my suggestion as much thought as I should have!

jkbullard:
Commit d80af4f0 allows .xibs to be built with Xcode 7.3.1.

I've also made several other commits to enhance the window's appearance or fix problems.

macbar:
Is there a way to disable this Security code field? In order to maintain compatibility with other clients, we need to stick with the old setup where a separate token popup is shown upon successful password authentication. This security code tricks our users into thinking they need to enter their token which makes the password authentication fail.

Thanks

macmoonshine:
It would be possible to hide the checkbox and the input field via a new boolean in the defaults. Would that be sufficient?

jkbullard:
> It would be possible to hide the checkbox and the input field via a new boolean in the defaults. Would that be sufficient?

That's what I meant by "set a per-configuration preference to enable this".

macbar:
That would be great, indeed!

macmoonshine:
I could add the setting and hiding already in the next days, so the UI elements could be hidden via `defaults write`. A customization of the settings dialog will probably take longer already because of the localization.
@jkbullard Do you think that an adjustment of the settings dialog is necessary?

jkbullard:
@macmoonshine wrote:
> @jkbullard Do you think that an adjustment of the settings dialog is necessary?

No, and I wouldn't even want to put it on the "Advanced" settings. (Which is what I meant when I wrote "it isn't a common enough thing to create a checkbox for".)

A PR for this would be very welcome!

macmoonshine:
I have created a new [pull request](https://github.com/Tunnelblick/Tunnelblick/pull/708).

jkbullard:
The pull request was accepted, and recently-released Tunnelblick 3.8.8beta03 includes support for hiding the checkbox.

PR #675, opened 2021-04-01
Update embedded OpenSSL to v1.1.1k

grumpybozo:
1.1.1k fixes 2 high-severity CVEs: CVE-2021-3449 and CVE-2021-3450 . See https://www.openssl.org/news/vulnerabilities-1.1.1.html


jkbullard:
This is in process; new stable and beta versions of Tunnelblick will be released within the next few days. possibly as early as today.

jkbullard:
Thank you, but this PR won't work because it doesn't modify third_party/Makefile. So OpenVPN 1.1.1j will be used. It also doesn't remove the 1.1.1j .zip.

See third_party/README.txt for instructions on updating Tunnelblick's third-party components.

It's also a duplicate of commit 7b58c1ba8b0b0c6830269bedc5af8dcdbff4c872.

PR #694, opened 2021-08-19
Set initial focus to security token if the password is filled

TWHOI:


jkbullard:
@TWHOI - Thanks! I want to think about whether it makes sense to do this if the checkbox for the security token is not checked, but I'm inclined to think it does, or that it doesn't matter. I'll play around with it.

jkbullard:
Thanks!

PR #708, opened 2021-12-19
Setting <VPN name>-loginWindowSecurityTokenIsHidden added to hide the…

macmoonshine:
… ui elements for the security token.

We're happy to accept Pull Requests (PRs) for everything **except translations**.

We do translations on Crowdin.com, where they can be voted up or down by other translators.
Then we import translations with the highest vote totals from Crowdin into the Tunnelblick
source code. It's a one-way transfer from Crowdin into the source code – we don't transfer
from the source code into Crowdin, so we don't accept PRs.

Please see [Localizing and Translating Tunnelblick](https://tunnelblick.net/cLocalizeTranslate.html) for details.

Email developers@tunnelblick.net to volunteer to translate.

### DIFF
--- a/tunnelblick/LoginWindow-RTL.xib
+++ b/tunnelblick/LoginWindow-RTL.xib
@@ -17,6 +17,7 @@
                 <outlet property="saveUsernameInKeychainCheckbox" destination="42" id="89"/>
                 <outlet property="securityEyeButton" destination="LS4-oh-7b6" id="8Tl-Hg-kPp"/>
                 <outlet property="securityToken" destination="adg-1U-zFx" id="w7u-hq-5y5"/>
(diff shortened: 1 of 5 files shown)


jkbullard:
Many thanks, @macmoonshine. Please also add the new preference name ("-") to defines.h in the definition of CONFIGURATIONS_PREFERENCES_NSARRAY.

(Note that the comma and the backslash in the line are critical.)

macbar:
Thanks both!
Any idea when a build containing this change will be released? 
I'm trying to assess whether or not to setup a build environment or just wait a few more days before the next beta release.

jkbullard:
I'll probably release a beta containing this change within the next few days.

jkbullard:
Recently-released Tunnelblick 3.8.8beta03 includes support for hiding the checkbox.

PR #712, opened 2022-01-24
Updated successful IP change message

GijsVermarien:
I noticed that the successfully connected 5 time message asks for donations, but the contributing page no longer does this, so I updated the message to reflect this. I am not sure whether this pull request is the best way of introducing this change. If there is another way I can contribute this, please let me know.


jkbullard:
@GijsVermarien - Thanks. Your work is very much appreciated, however, we cannot accept changes to .strings file via pull requests. That's because our translation flow is highly automated and one way: from our source code, to Crowdin (the platform on which we do translations), and then back to the our source code.

All that is necessary to do to change a string is to change it in the Objective-C source code (the .m file).

Our automated process will integrate the change into the en.lproj/Locaizable.strings file, commit that change, and send the change to Crowdin, where our translators translate it. When we do a release, another automated process integrates the translations back into the Localizable.strings filesand thencommits the changes in the other Localizable.strings files before building binaries from the source code.

So I'm sorry, but I must reject this PR. A PR that changes only the string in VPNConnection.m would be much appreciated.

GijsVermarien:
I made a mistake in reverting, I am opening another PR instead.

PR #713, opened 2022-01-24
Updated successful IP change message

GijsVermarien:
I made a mistake with reverting in #712, so I created this new pull request with only the .m file altered. This update reflects the messaging on the website.


jkbullard:
@GijsVermarien - Thanks! Looks great!

PR #719, opened 2022-03-30
Update free public dns server list to include public Comcast Xfinity DNS servers

zNeill:
Request to add the Comcast/Xfinity public DNS servers to the recognized list of Free Public DNS servers. Small one-line change.


jkbullard:
Is there a Comcast/Xfinity page that confirms that they are **public** DNS servers? Usually ISP's DNS servers only work from within their network, so they aren't public.

zNeill:
@jkbullard you're correct, they aren't public, my apologies for the erroneous request.

PR #724, opened 2022-04-21
Support native macOS IOUserEthernet devices for TAP

sbingner:
This patch uses IOUserEthernet to support TAP tunnels on macOS. 

This also changes the default behavior for tap kext to not be to always try to load it.

It fetches the missing header from apple directly since the header is listed as APSL to avoid conflicts with GPL.

I'm happy to make or accept changes or work with you to clean up anything necessary.  I will be submitting the patch upstream but I suspect this might be a good place to get it ironed out more first, and you're welcome to do the upstream submission as well.

The patch is not yet applied to the oldest version of openvpn bundled.  It works with the two newer versions.

### DIFF
--- a/third_party/makefiles/Makefile-openvpn
+++ b/third_party/makefiles/Makefile-openvpn
@@ -34,6 +34,7 @@ built-openvpn-prepare:
 		openvpn_name="$${openvpn_tar_name%.*}" ; \
 		mv $$openvpn_name   openvpn ; \
 		cd  openvpn ; \
+		curl -o IOUserEthernetController.h https://opensource.apple.com/source/IOKitUser/IOKitUser-1845.120.6/network.subproj/IOUserEthernetController.h; \
 		if [ -d $(OPENVPN_SOURCE_DIR)/$$openvpn_folder_name/patches ]; then \
 			for patch_file in $$(ls $(OPENVPN_SOURCE_DIR)/$$openvpn_folder_name/patches/*.diff); do \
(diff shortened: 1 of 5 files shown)


sbingner:
I fixed the leaks you pointed out and one other one... I actually did talk to OpenVPN about this before I made this patch, AFAIK what you suggest would be fine but they liked the curl method better because it made it clear that the headers were system headers... AFAIK you are correct about it being OK to put the few (one?) missing headers in, but I'm not quite sure and I'm not a lawyer.  If I hadn't looked at the headers already I could probably have just created headers of my own and been fine.  Maybe somebody else who hasn't looked at them wants to generate headers based off looking at how they are used in this code without looking at any apple code?  That, IMO, would be the most legal way to get the headers in.

jkbullard:
Thanks for fixing the leaks.

I'm OK with using the "curl" method for including the header file as long as OpenVPN will accept it.

Would it be reasonable to create the folder and use "curl" only if the folder and file file don't already exist? If you did that I could more easily incorporate this patch into my release build process.

I may have more comments on the code once I've looked at it more closely

sbingner:
We could… or just if it doesn’t exist in search path, I have configure checking for it already, it could probably just do curl if it failed to find it

jkbullard:
@sbingner - Hi! Thanks again for this patch!

Are you following the discussion about a possible problem with this?

It's taking place at  https://groups.google.com/g/tunnelblick-discuss/c/HbM3lR517oU

Issue #676, opened 2021-04-07
RFE: Tunnelblick support for openvpn v2.5 pending authentication


pasikarkkainen:
Openvpn v2.5 introduced new feature "pending authentication". It'd be nice to have Tunnelblick support these new features.

Quote from openvpn Changes file ( https://github.com/OpenVPN/openvpn/blob/release/2.5/Changes.rst )

```
Improved support for pending authentication

    The protocol has been enhanced to be able to signal that the authentication should use a secondary authentication via web (like SAML) or a two factor authentication without disconnecting the OpenVPN session with AUTH_FAILED. The session will instead be stay in a authenticated state and wait for the second factor authentication to complete.

    This feature currently requires usage of the managent interface on both client and server side. See the management-notes.txt "client-pending-auth" and "cr-response" commands for more details.
```

quote(s) from https://github.com/OpenVPN/openvpn/blob/master/doc/management-notes.txt

"client-pending-auth":
```
COMMAND -- client-pending-auth  (OpenVPN 2.5 or higher)
----------------------------------------------------

Instruct OpenVPN server to send AUTH_PENDING and INFO_PRE message
to signal a pending authenticating to the client. A pending auth means
that the connecting requires extra authentication like a one time
password or doing a single sign one via web.

    client-pending-auth {CID} {EXTRA} {TIMEOUT}

The server will send AUTH_PENDING and INFO_PRE,{EXTRA} to the client. If the
client supports accepting keywords to AUTH_PENDING (announced via IV_PROTO),
TIMEOUT parameter will be also be announced to the client to allow it to modify
its own timeout. The client is expected to inform the user that authentication
is pending and display the extra information and also show the user the
remaining time to complete the auth if applicable.

Receiving an AUTH_PENDING message will make the client change its timeout to
the timeout proposed by the server, even if the timeout is shorter.
If the client does not receive a packet from the server for hand-window the
connection times out regardless of the timeout. This ensures that the connection
still times out relatively quickly in case of network problems. The client will
continously send PULL_REQUEST messages to the server until the timeout is reached.
This message also triggers an ACK message from the server that resets the
hand-window based timeout.

Both client and server limit the maximum timeout to the smaller value of half the
--tls-reneg minimum time and --hand-window time (defaults to 60s).

For the format of EXTRA see below. For the OpenVPN server this is a stateless
operation and needs to be followed by a client-deny/client-auth[-nt] command
(that is the result of the out of band authentication).

[...]

PR #766, opened 2023-09-15
Added support for client-pending-auth with WEB_AUTH

maurice2k:
Adds support for client-pending-auth with WEB_AUTH.

See https://github.com/Tunnelblick/Tunnelblick/issues/676.

### DIFF
--- a/tunnelblick/VPNConnection.m
+++ b/tunnelblick/VPNConnection.m
@@ -4472,6 +4472,39 @@ static pthread_mutex_t lastStateMutex = PTHREAD_MUTEX_INITIALIZER;
         [self sendStringToManagementSocket: @"hold release\r\n" encoding: NSASCIIStringEncoding];
         return;
     }
+
+    if (  [line hasPrefix: @">INFOMSG:WEB_AUTH:"]  ) {
+
+        NSString *pattern = @"^>INFOMSG:WEB_AUTH:([^:]*):(.+)";
+        NSError *error = nil;
+
+        NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:pattern options:0 error:&error];
+        if (error) {
+            NSLog(@"Error creating regular expression: %@", error);
+            return;
+        }
+
+        NSRange range = NSMakeRange(0, line.length);
+        NSTextCheckingResult *match = [regex firstMatchInString:line options:0 range:range];
+
+        if (match) {
+            NSString *urlString = [line substringWithRange:[match rangeAtIndex:2]];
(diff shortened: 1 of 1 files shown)


jkbullard:
Thanks!

Although this code ignores the flags, I assume it's useful for some situations, and dealing with the flags could be added later.

I think one change is needed: add a line with `return;` before the closing `}` at (new) line 4507. That will avoid dropping through to the existing code.

maurice2k:
I've added the return statement and a comment that flags are currently not supported.
I had a look at the only other GUI I'm aware of that is supporting this (OpenVPN Connect) and they also just ignore the flags: https://github.com/OpenVPN/openvpn-gui/blob/7de26e65af2252d66114a0adadfd3136161c086b/openvpn.c#L1635

jkbullard:
@maurice2k - Thanks again!

schwabe:
That's nice. BUt I think to really use this Tunnelblick should add `--setenv IV_SSO webauth` to the commandline since otherwise you have to manually add that to configs.

maurice2k:
> That's nice. BUt I think to really use this Tunnelblick should add `--setenv IV_SSO webauth` to the commandline since otherwise you have to manually add that to configs.

Good point, only had it in my config files. Added it in another PR: https://github.com/Tunnelblick/Tunnelblick/pull/767

maurice2k:
> @maurice2k - Thanks again!

You're welcome and thanks for creating/maintaining TunnelblicK!

tunnelblick/VPNConnection.m
```diff
@@ -4472,6 +4472,41 @@ -(void) processRealTimeOpenvpnOutputWithLine: (NSString *) line {
         [self sendStringToManagementSocket: @"hold release\r\n" encoding: NSASCIIStringEncoding];
         return;
     }
+
+    if (  [line hasPrefix: @">INFOMSG:WEB_AUTH:"]  ) {
+        // This is currently not supporting any flags; it just opens the given URL
+        // in the defauklt web browser.
```
schwabe, line 4478: Minor typo in default web browser.

schwabe:
@maurice2k openvpn3-linux and OpenVPN for Android also support webauth method. OpeNVPN3-linux also ignores the flags. OpenVPN for Android actually implement an internal webview for the internal flag support and the javascript support from the spec for that but after OpenVPN Cloud/Connexa does not use that internal webview mode anymore, there is nobody left using that, so ignoring the flags should be fine for now.

tunnelblick/VPNConnection.m
jkbullard: Thanks, @schwabe; I've fixed that

jkbullard:
Thanks for all the help, @schwabe.

PR #790, opened 2024-03-11
Fix case of AppKit

ryandesign:
Fixes build on case-sensitive filesystems.

### DIFF
--- a/tunnelblick/ConfigurationManager.h
+++ b/tunnelblick/ConfigurationManager.h
@@ -29,7 +29,7 @@
  */
 
 #import "defines.h"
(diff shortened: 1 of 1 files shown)


ryandesign:
> Several other folders have upper-case names, so I assume they may cause a problem, too.

The other folders seem to be referred to by the correct case in the Xcode project.

> As to the error about 54 .lproj folders, that might be corrected if you do Product >> Clean Build Folder before building.

Thanks, I will try again later. I did look in the project's Resources, to the extent that I understand what they mean by that, and there sure seemed to be more than 12 of them there.

PR #827, opened 2024-11-17
Update tunnelblick-uninstaller.sh to fix unicode not equal

mcoliver:
should be != not the unicode ≠


jkbullard:
@mcoliver - Thanks for catching this!

PR #828, opened 2024-11-25
Announce crtext support, too

egroeper:
Tunnelblick supports crtext for quite some time. It works fine. We should announce that, too.
Then servers could select the method they support.

Now: IV_SSO=webauth
Then: IV_SSO=webauth,crtext

We are using crtext with Tunnelblick in production. But currently we need a special solution for Tunnelblick, because it doesn't announce its support for crtext. We don't support webauth on server side.

Extends #767

### DIFF
--- a/tunnelblick/tunnelblick-helper.m
+++ b/tunnelblick/tunnelblick-helper.m
@@ -2640,7 +2640,7 @@ int startVPN(NSString * configFile,
 
     [arguments addObject: @"--setenv"];
     [arguments addObject: @"IV_SSO"];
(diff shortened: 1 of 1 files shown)


jkbullard:
Thanks!

hanzhinstas:
Sorry, but aren't you mixing up `>INFOMSG:CR_TEXT:<flags>:<message>` with `CRV1`? I don't see that Tunnelblick code has any support for popping out challenge-response support when management server issues that type of INFOMSG

PR #856, opened 2025-06-25
Adds support for client-pending-auth using INFOMSG:CR_TEXT request

hanzhinstas:
See Tunnelblick#676 for protocol details. This change is a simple implementation and has some TODO left as I'm not an Objective-C developer at all.

Tested locally.

### DIFF
--- a/tunnelblick/VPNConnection.m
+++ b/tunnelblick/VPNConnection.m
@@ -4590,6 +4590,48 @@ static pthread_mutex_t lastStateMutex = PTHREAD_MUTEX_INITIALIZER;
         return;
     }
 
+    if (  [line hasPrefix: @">INFOMSG:CR_TEXT:"]  ) {
+        // This is currently not supporting any flags; it just opens the given URL
+        // in the default web browser.
+        NSString *pattern = @"^>INFOMSG:CR_TEXT:([^:]*):(.+)";
+        NSError *error = nil;
+
+        NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:pattern options:0 error:&error];
+        if (error) {
+            NSLog(@"Error creating regular expression: %@", error);
+            return;
+        }
+
+        NSRange range = NSMakeRange(0, line.length);
+        NSTextCheckingResult *match = [regex firstMatchInString:line options:0 range:range];
+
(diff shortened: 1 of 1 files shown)


jkbullard:
Thanks!

I realize you modeled this after PR #766, which uses a regular expression (regex), but after the ">INFOMSG:CR_TEXT:", there are only two fields, separated by a colon: the flags and the prompt text. Parsing the line something like the following avoids a regex:
```
    NSRange r = NSMakeRange(@">INFOMSG:CR_TEXT:".length, line.length - @">INFOMSG:CR_TEXT:".length);
    NSArray items = [[line substringWithRange: r] componentsSeparatedByString: @":"];
    if (  items.count != 2. ) {
        NSLog(@"Syntax error in management command '%@'", line);
        return;
    }
    NSString * promptString = [items objectAtIndex: 0];
    NSString * flagsString = [items objectAtIndex: 1];
```
 4. The flag arguments for `getResponseToChallenge:echoResponse:responseRequired:isStatic:` are then easily set from flagsString with something like:
```
    NSArray * flags = [flagsString componentsSeparatedByString @","];
    BOOL echo = [flags containsObject: @"E"];
    BOOL required = [flags containsObject: @"R"];
```
5. Suggestion (if a regex is still used): add the contents of the `line` variable in the error message to help debugging if the regex doesn't match.
6. Suggestion: break up line 4627: create an NSString with the base64-encoded response in one line, and send it to the socket in a second line.

hanzhinstas:
I would also suggest refactoring getResponseToChallenge in a separate PR as isStatic is something from other 2fa flow that assumes you have some scripts for either static or dynamic CR. Here we don't probably need this (and all the filesystem lookups too).

jkbullard:
>  I'm ready to patch dropping these quotes.

I don't think that's necessary at this point. If it works for you with the double-quotes, I think they can be left in.

My comment was for the record in case it causes a problem in the future.

hanzhinstas:
@jkbullard may I ask you to include this to 8.1? We really need this as OpenVPNConnect that supports CR_TEXT may in rare cases get in conflict with Tunnelblick and AnyConnect (race conditions) and we suffer from this

PR #871, opened 2025-11-28
External certificate and signing facility support

ne-bknn:
Hi!

About 6 month ago I asked about Tunnelblick support for Secure Enclave-backed keys on the discussion group: https://groups.google.com/g/tunnelblick-discuss/c/9ymj0RJLdjM/m/Ayrum2PjAgAJ, via the management interface's external-cert and external-key features.

I agree with your position that my initial proposal was too big and ambitious, and your alternative is perfectly reasonable. And I am grateful that you offered a viable alternative instead of rejecting right away. Unfortunately I was pulled elsewhere in my company, thus the ghosting)

I've implemented the signing server in Swift (making last adjustments to open the repo, will post here in a few days)...

...and LLM-generated your vision of the feature into Tunnelblick, just as the proof of concept and to test my side of things. Feature spec, if you like. I have some experience with Swift and lots with pure C, but none with Objective-C. For me the patch seems all right, but I do not expect you to merge this; this may be very bad and I would have no idea.

So, the question is, what needs to happen for this feature (mine or one rewritten from scratch) to land in Tunnelblick?

### DIFF
--- a/tunnelblick/VPNConnection.h
+++ b/tunnelblick/VPNConnection.h
@@ -164,6 +164,10 @@ struct Statistics {
 
 	BOOL volatile   skipConfigurationUpdateCheckOnce; // True only after have skipped a configuration update, so the next
 													  // connection attempt will not try to check for the update
+    
+    // External key proxy support (for --management-external-key and --management-external-cert)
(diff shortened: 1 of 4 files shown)


ne-bknn:
I've released the signing server: https://github.com/ne-bknn/nailed

PR #900, opened 2026-07-05
Clarify 'Display connection timers' info text

MiMoHo:
### What this PR now does

Following the maintainer's feedback below, this PR no longer changes any behavior. As suggested, it now only revises the pop-up description of the "Display connection timers" checkbox (Preferences → Appearance) so that it describes what the timer shows in each phase:

- while a connection attempt is being made: the time that has elapsed trying to make the connection,
- when a connection is made: the timer restarts and shows the time that has elapsed since the connection was made,
- while a disconnect attempt is being made: the timer restarts and shows the time that has elapsed trying to disconnect.

The previous description only mentioned "the time since a connection was established", which is what led to this PR in the first place.

The original behavior-change commit has been reverted in-branch; the net diff against `main` is the info-text change in `AppearanceView.m` only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tunnelblick/AppearanceView.m
+++ b/tunnelblick/AppearanceView.m
@@ -139,7 +139,10 @@ TBSYNTHESIZE_OBJECT_GET(retain, TBButton *,        appearanceDisplayStatisticsWi
 	
 	[appearanceDisplayConnectionTimersCheckbox
 	  setTitle: NSLocalizedString(@"Display connection timers", @"Checkbox name")
-	 infoTitle: attributedStringFromHTML(NSLocalizedString(@"<p><strong>When checked</strong>, the time since a connection was established is displayed in the Tunnelblick menu and in the title of the VPN Details window.</p>",
+	 infoTitle: attributedStringFromHTML(NSLocalizedString(@"<p><strong>When checked</strong>, a timer is displayed in the Tunnelblick menu and in the title of the VPN Details window:</p>\n"
+														   @"<p>While a connection attempt is being made, the timer displays the time that has elapsed trying to make the connection."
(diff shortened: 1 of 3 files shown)


jkbullard:
Thanks, but I don't think I want to change the current behavior.

Currently, while a connection attempt is being made, the timer contains and the UI displays the time that has elapsed trying to make a connection. Then, when a connection is made, it shows the time that has elapsed since the connection. Finally, while a disconnect attempt is being made, the timer shows the time trying to disconnect.

As I understand it, this PR would modify this behavior by showing 0:00 until a connection is made. I don't think that's good; it displays _less_ information. It is sometimes helpful to see how long it takes to make the connection, and how long it takes to disconnect, and this PR would make that more difficult.

I admit that the variable names "connectedSince", "connectedSinceDate", etc., are a bit misleading. They could be renamed to be "connectedOrConnectingOrDisconnectingSince...", but that's a bit unwieldy.

This does point out that the pop-up description of the "Display connection timers" checkbox could be revised to describe more fully what the timer shows while connecting and disconnecting.

MiMoHo:
Thanks for the explanation — understood and agreed: seeing how long connect and disconnect attempts take is useful information, and the behavior should stay as it is.

I have reworked this PR accordingly: the behavior change is reverted, and the PR now only revises the pop-up description of the "Display connection timers" checkbox, as you suggested, with wording based on your description above. The net diff is now a single string in `AppearanceView.m`. Happy to adjust the wording — or to close this PR if you would rather handle it another way.

jkbullard:
Thanks, the wording looks fine. It would be nice if you break the last paragraph into three bullet points (an HTML unordered list); that would make it easier to read.

PR #903, opened 2026-08-11
Use a certificate and private key from the Keychain (management-external-cert/key)

Knut0:
### What this does

On Windows a configuration can keep the certificate and the private key in the system store and refer to them with `cryptoapicert "SUBJ:…"`. On macOS there is no equivalent: `cryptoapicert` is Windows-only (and Tunnelblick rejects it), and the OpenVPN binaries that Tunnelblick includes are built with `enable_pkcs11=no`, so PKCS#11 is not available either. That leaves keeping the certificate and the private key in a file inside the configuration.

This implements the remaining alternative: OpenVPN's `--management-external-cert` and `--management-external-key`, answered by Tunnelblick as the management client. The certificate and the private key stay in the user's login keychain, and the configuration contains only a selector:

```
management-external-cert "SUBJ:CN=my name"
management-external-key pkcs1 pss nopadding
```

Both options are already in `OPENVPN_OPTIONS_THAT_ARE_SAFE`, so nothing about configuration validation changes. What was missing was handling of the `>NEED-CERTIFICATE` and `>PK_SIGN` messages.

### How it works

* When the management socket is connected and the configuration includes `management-external-key`, Tunnelblick sends `version 3` before `pid`. Version 3 is what makes OpenVPN include the signature algorithm in `>PK_SIGN`; without it OpenVPN uses the obsolete `>RSA_SIGN`, which carries no algorithm and cannot be used with TLS 1.3. It is sent *only* for those configurations, and it is version 3 rather than something higher on purpose: `MCV_USERNAME_ONLY` (4) and `MCV_MULTILINE_PASSWORD` (5) change the format of the username and password requests, which would change behavior for every other configuration.
* `>NEED-CERTIFICATE:<selector>` → find the identity in the Keychain and answer with the certificate as PEM. The answer has to be a full PEM including the armor lines, not the bare base64 that `doc/management-notes.txt` describes: OpenVPN hands it to `tls_ctx_load_cert_file(…, inline = true)`, i.e. to the PEM parser. Only the certificate itself is sent; an intermediate CA, if there is one, goes into the configuration as `<extra-certs>`.
* `>PK_SIGN:<base64>[,<algorithm>]`, and the legacy `>RSA_SIGN:`, → sign with `SecKeyCreateSignature` and answer with `pk-sig` (`rsa-sig` for the legacy message). Signing is done off the main thread, because using the private key may present a Keychain dialog and wait for the user; the response is written to the socket back on the main thread. The base64 of a signature made with a key of up to 6144 bits fits in the 1024-byte line buffer OpenVPN reads the response with, so it is sent as a single line.
* If anything fails — no matching certificate, an algorithm macOS cannot produce, the user denies access to the key — the reason goes to the log and to an alert, and OpenVPN is sent a SIGTERM instead of being left waiting for a response it will never get.
* The identity, and with it the reference to the private key, is released when the management socket goes away.

### The selector

The syntax mirrors `cryptoapicert`, so a configuration can be moved between Windows and macOS by changing only the option name:

| Selector | Matches |
| --- | --- |
| `SUBJ:<string>` | certificate whose subject DN contains `<string>` (case-insensitive) |
| `ISSUER:<string>` | certificate whose issuer DN contains `<string>` (case-insensitive) |
| `THUMB:<hex>` | certificate whose SHA-1 fingerprint is `<hex>` (spaces and colons ignored) |
| `<string>` | the same as `SUBJ:<string>` |

Certificates that are not valid at the current time are skipped. If several identities match, the one with the latest `notBefore` is used and all of the matches are written to the log, so the ambiguity is visible. There is no GUI for choosing a certificate.

Signature algorithms: `RSA_PKCS1_PADDING`, `RSA_NO_PADDING`, `ECDSA`, and `RSA_PKCS1_PSS_PADDING` with `hashalg=SHA1/SHA224/SHA256/SHA384/SHA512`. `saltlen=max` is refused with an explanatory message, because the Security framework can only produce a salt as long as the digest — which is what OpenSSL asks for in TLS 1.3, so in practice it does not come up.

### Files

* `tunnelblick/TBKeychainIdentity.{h,m}` (new) — everything that touches the Keychain and the crypto, with no UI dependencies. Every `Sec*`/`CF*` result is checked and every CF object is released on every path.
* `tunnelblick/VPNConnection.{h,m}` — the conditional `version 3`, two new message handlers, one instance variable, and releasing the identity when the connection ends.
* `tunnelblick/Tunnelblick.xcodeproj/project.pbxproj` — the new files, in the Tunnelblick target only. tunnelblick-helper and openvpnstart do not need them: the Keychain is reached from the user's GUI process.
* `tunnelblick/TBKeychainIdentityTest.m` (new, in a separate commit) — a small command line program that exercises `TBKeychainIdentity` against the Keychain of whoever runs it, without connecting to a VPN. It is not part of any target and is built with the single clang command in its header comment. Happy to drop that commit if you would rather not carry it.

### Testing

Tested on macOS 26.5.2 with Tunnelblick built from 9.0beta06 plus this change, using the OpenVPN that Tunnelblick includes, against an OpenVPN 2.7 server that requires `tls-version-min 1.3`:

* an RSA identity in the login keychain, selected with `management-external-cert "SUBJ:CN=…"`: OpenVPN asked for a PSS signature, the handshake completed and the tunnel came up, with `>NEED-CERTIFICATE` and `>PK_SIGN` and the answers to them visible in the log;
* a TLS renegotiation in the middle of that session: OpenVPN asked for another signature on the same connection, it was created from the identity that was still held for the connection, and the tunnel stayed up;
* the command line program from the second commit: `SUBJ:`, `ISSUER:` and `THUMB:` matching, including a fingerprint written with colons and in mixed case, an empty selector and a selector that matches nothing, plus a signature created and then verified with the certificate's public key for each of the algorithms above;
* configurations that do not use `management-external-key` are unaffected: `version 3` is not sent for them, so the management dialogue for them is unchanged.

### DIFF
--- a/tunnelblick/TBKeychainIdentity.h
+++ b/tunnelblick/TBKeychainIdentity.h
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2026 Andrew Grishenko. All rights reserved.
+ *
+ *  This file is part of Tunnelblick.
+ *
+ *  Tunnelblick is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License version 2
+ *  as published by the Free Software Foundation.
+ *
+ *  Tunnelblick is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program (see the file COPYING included with this
+ *  distribution); if not, see http://www.gnu.org/licenses/.
+ */
+
+#import <Foundation/Foundation.h>
+#import <Security/Security.h>
+
+// Finds identities (a certificate together with its private key) in the user's Keychain and signs
+// data with them, so that OpenVPN's "--management-external-cert" and "--management-external-key"
+// options can be used on macOS instead of the Windows-only "--cryptoapicert" option.
+//
+// The argument to "--management-external-cert" is passed through to the management interface in the
+// ">NEED-CERTIFICATE:" message and is used here to select which identity to use. The syntax mirrors
+// the syntax of "--cryptoapicert":
+//
+//      SUBJ:<string>   Use the identity whose certificate's subject contains <string>
+//      ISSUER:<string> Use the identity whose certificate's issuer contains <string>
+//      THUMB:<string>  Use the identity whose certificate's SHA-1 fingerprint is <string>
+//
+// A selector without one of those prefixes is interpreted as if it had the "SUBJ:" prefix.
+
+@interface TBKeychainIdentity : NSObject
+
+// Returns a retained SecIdentityRef which the caller must CFRelease(), or NULL if an identity could
+// not be found (in which case *errMsgPtr is set to a localized message describing the problem).
++(SecIdentityRef) copyIdentityMatchingSelector: (NSString *)  selector
+                                  errorMessage: (NSString **) errMsgPtr;
+
+// Returns the identity's certificate as PEM (including the "-----BEGIN CERTIFICATE-----" and
+// "-----END CERTIFICATE-----" lines, separated by LF characters, without a trailing LF), or nil if
+// it could not be obtained (in which case *errMsgPtr is set to a localized message).
++(NSString *) pemForIdentity: (SecIdentityRef) identity
+                errorMessage: (NSString **)    errMsgPtr;
+
+// Signs data with the identity's private key and returns the signature, or nil if the data could
+// not be signed (in which case *errMsgPtr is set to a localized message).
+//
+// 'algorithm' is the algorithm string from the management interface's ">PK_SIGN:" message (that is,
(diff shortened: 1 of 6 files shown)


Knut0:
Hi — a gentle nudge on this one, no urgency intended.

Quick reminder of what it is: #903 adds handling of `>NEED-CERTIFICATE` and
`>PK_SIGN`, so a configuration can use a certificate and private key kept in the
Keychain via `management-external-cert`/`management-external-key` — the macOS
counterpart of what `cryptoapicert` does on Windows.

I realise the last few weeks went into 9.0 and 10.0beta01, so this may simply not
have reached the top of the pile yet. The branch still merges cleanly into `main`;
I am happy to rebase it onto 10.0beta01 if that is more convenient for you.

Before you spend time on the code, the more useful question is probably whether
this is a feature you want in Tunnelblick at all. If the direction is wrong, I would
much rather hear that than leave the PR sitting open. If it is right but the shape
is not, I am glad to rework it — split the change up, drop the second commit with
the command line test program (it is a separate commit for exactly that reason), or
anything else you would like changed.

Thank you for your work on Tunnelblick, and for taking a look whenever you get to it.

PR #904, opened 2026-08-13
fix: harden tunnelblickd accept-loop error handling

SebTardif:
## Summary

Hardens `tunnelblickd`'s accept loop so bad clients do not leak file
descriptors, crash the daemon, or take it down via `goto done`.

## Problem

After `accept()`, several error paths used `continue` without
`close(filedesc)`, so failed clients leaked sockets until the 100-event
process recycle.

`getpwuid(client_euid)` was used without a NULL check, so a peer UID with
no passwd entry could crash the root daemon.

Username/home UTF-8 failures used `goto done`, which exits the whole
daemon instead of dropping that connection.

Read errors were logged with `Error from read(): &m`, so errno was never
expanded (`%m` is the asl/syslog form used elsewhere in this file).

## Change

- Close the client FD on every post-accept error path before `continue`
- NULL-check `getpwuid`; on failure, log, close, continue
- UTF-8 username/home failures: close and continue (not `goto done`)
- Fix read-error log format: `&m` -> `%m`

## Validation

- Source inspection of the accept loop (all post-accept continues close)
- Host checks: format string does not expand `&m`; NULL passwd path is handled
- Full Xcode build of Tunnelblick not run in this environment (third_party /
  notarized app build)

## References

- Accept-loop pattern from 33223141 (2015, launchd daemon era)
- getpwuid path refined in 945fca65 (2020)

### DIFF
--- a/tunnelblick/tunnelblickd.m
+++ b/tunnelblick/tunnelblickd.m
@@ -752,7 +752,8 @@ int main(void) {
 		if (  0 != getpeereid(filedesc, &client_euid, &client_egid)  ) {
 			asl_log(asl, log_msg, ASL_LEVEL_ERR, "Could not obtain peer credentials from unix domain socket: %m; our uid = %lu; our euid = %lu; our gid = %lu; our egid = %lu",
 					(unsigned long)getuid(), (unsigned long)geteuid(), (unsigned long)getgid(), (unsigned long)getegid());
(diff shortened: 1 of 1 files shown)


jkbullard:
Thanks! This looks good at first glance. It will be a while (possibly weeks?) before I accept it into the main branch; I'm pretty busy.

PR #908, opened 2026-08-23
fix: apply config-name length limit to argv[2]

SebTardif:
Apply the config-name length limit to the actual config argument.

## Problem

The `start` command stored `argv[2]` (the config name) only when `strlen(argv[3])` was under `DISPLAY_NAME_LENGTH_MAX`. `argv[3]` is the port, which is always short when valid, so the name-length gate never ran. `validateConfigName` also did not enforce the 512-character maximum.

## Change

- Assign the config name from `argv[2]` when it is valid UTF-8.
- Reject names longer than `DISPLAY_NAME_LENGTH_MAX` in `validateConfigName` (exit 175).

## Validation

- Traced `start` argument order: config, port, useScripts, ...
- A too-long name now fails validation instead of being ignored while start proceeds with the default `"X"`.
- No helper unit-test harness (same as #904).

## Related

### DIFF
--- a/tunnelblick/tunnelblick-helper.m
+++ b/tunnelblick/tunnelblick-helper.m
@@ -3223,6 +3223,10 @@ static void validateConfigName(NSString * name) {
         Log(@"Configuration name is empty");
         exitOpenvpnstart(172);
     }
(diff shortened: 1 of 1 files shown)


jkbullard:
Thanks!

PR #909, opened 2026-08-23
fix: avoid Keychain CFRelease(NULL) and atomic config rewrite

SebTardif:
Avoid a Keychain logging crash and do not truncate a live config before a successful write.

## Problem

1. `KeyChain` `logStatus:` always `CFRelease`s `SecCopyErrorMessageString`. That API may return NULL. `CFRelease(NULL)` is documented to crash. The method runs on retrieve, add, and delete failures.
2. Path-stripping in `ConfigurationConverter` opened the live `.ovpn` with `fopen(..., "w")`, which truncates immediately. A failed `fwrite` left an empty file.

## Change

- Release the error string only when it is non-NULL.
- Write the modified config with `writeToFile:atomically:YES`, matching the other copy paths in the same file.

## Validation

- Apple documents `SecCopyErrorMessageString` as returning NULL when no string is available.
- `writeToFile:atomically:YES` is already used at ConfigurationConverter.m:566 and :575.
- No unit-test harness for these GUI paths (same as #904).

## Related


jkbullard:
Thanks!

PR #910, opened 2026-08-23
fix: bound helper and tunnelblickd NSTask waits

SebTardif:
Give helper and tunnelblickd NSTask waits a deadline.

## Problem

`tunnelblickd` is a single-threaded accept loop. It waits with `[task waitUntilExit]` for every helper command. The helper `runAsRoot` path does the same for scripts, `kextunload`, `killall`, and `codesign`. A hung child freezes all further VPN IPC until the process is killed.

## Change

Use the existing `startTool` pattern in `sharedRoutines.m` (warn, then SIGTERM):

- Helper children: warn every 10s, terminate after 60s (same as `startTool`).
- Daemon waiting on the helper: warn every 30s, terminate after 300s (helper commands include app updates).

`becomeRoot` still runs after the wait, so the daemon restores root even if the helper is terminated.

## Validation

- Copied the wait loop from `startTool` (`sharedRoutines.m` 1386-1411).
- Confirmed `startVPN` returns after launching OpenVPN, so the 60s helper-child limit is for short tools, not the VPN session.
- No daemon unit-test harness (same as #904).

## Related

### DIFF
--- a/tunnelblick/tunnelblick-helper.m
+++ b/tunnelblick/tunnelblick-helper.m
@@ -1074,7 +1074,24 @@ static int runAsRootWithConfigNameAndLocCodeAndmanagementPasswordReturnOutput(NS
     becomeRoot([NSString stringWithFormat: @"launch %@", [thePath lastPathComponent]]);
 
         [task launch];
-        [task waitUntilExit];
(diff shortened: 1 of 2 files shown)


jkbullard:
Thanks!

PR #911, opened 2026-08-23
fix: escape easy-rsa path before embedding it in AppleScript

SebTardif:
Escape the easy-rsa folder path before embedding it in AppleScript.

## Problem

`openTerminalWithEasyRsaFolder` built `set cmd to "cd \"PATH\""` with no escaping. A path containing `"` or `\` could break out of that AppleScript string when the user clicks Utilities to open easy-rsa in Terminal.

The path comes from the `easy-rsaPath` preference or the default Application Support folder. This is hygiene for a user-controlled path, not a privilege-boundary issue.

## Change

- Escape `\`, `"`, CR, and LF for a double-quoted AppleScript string.
- Set `thePath` to that string and build `cd` with AppleScript `quoted form of`, so Terminal gets one shell token.

## Validation

- Backslash is escaped first so later quote escaping stays intact.
- `quoted form of` is the usual AppleScript way to produce a POSIX-safe token.
- No GUI unit-test harness (same as #904).

## Related

### DIFF
--- a/tunnelblick/easyRsa.m
+++ b/tunnelblick/easyRsa.m
@@ -427,6 +427,17 @@ BOOL secureOurEasyRsa(void) {
     return YES;
 }
 
+static NSString * appleScriptEscapedString(NSString * s) {
+
+    // Escape for a double-quoted AppleScript string.
+    NSMutableString * escaped = [[s mutableCopy] autorelease];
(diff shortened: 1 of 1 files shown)


jkbullard:
Thanks!

Issue #906, opened 2026-08-20
Menu bar icon position (NSStatusItem Preferred Position) is deleted on every launch


FrancoTampieri:
**Describe the bug**
Tunnelblick does not restore the menu bar icon position set by the user via Cmd+drag. On every launch, the saved `NSStatusItem Preferred Position Item-0` preference is deleted, so the icon always appears at the default position instead of where the user placed it.

**To Reproduce**
1. Cmd+drag the Tunnelblick menu bar icon to a custom position
2. Verify the position is saved: `defaults read net.tunnelblick.tunnelblick | grep "Preferred Position"` -> `"NSStatusItem Preferred Position Item-0" = 282;`
3. Quit and relaunch Tunnelblick: `killall Tunnelblick && sleep 2 && open -a Tunnelblick`
4. Read the preference again (also after waiting 10+ seconds to rule out cfprefsd caching): the `Preferred Position` key is gone; only `NSStatusItem VisibleCC Item-0 = 1` remains
5. The key only reappears after manually Cmd+dragging the icon again

**Expected behavior**
The saved icon position should be preserved across launches and the icon restored to the user-chosen position.

**Screenshots**
N/A (terminal output included above)

**Additional context**
On this same installation, changing the menu bar icon set caused Tunnelblick to crash when opening Settings. The status item handling code may be involved in both issues (crash report available on request).

**Diagnostic info**
- Tunnelblick version: 9.0beta09 (build 6480)
- macOS version: 26.6.2 (build 25G83)

jkbullard:
Thanks very much for reporting this, and for providing so much detail. It will be fixed in the next beta.

It isn't that Tunnelblick deletes the preference, it's that macOS deletes it when, as part of the creation of the status item, Tunnelblick sets its status to "hidden". (Hiding it temporarily was added while debugging other problems.) If we don't hide it, the preference isn't deleted and the icon appears in position it was Cmd+dragged to.

Please do send the crash report! You can email it to developers@tunnelblick.net if you don't want to post it publicly.

PR #912, opened 2026-08-24
fix: add missing comma in program-preference allowlist

SebTardif:
The program-preference allowlist is missing a comma after `NSStatusItem Preferred Position Item-0`, so that key and `NSWindow Frame SettingsSheetWindow` become one C string.

## Problem

Adjacent Objective-C string literals concatenate. In `NON_CONFIGURATIONS_PREFERENCES_NSARRAY`:

```
@"NSStatusItem Preferred Position Item-0" \
@"NSWindow Frame SettingsSheetWindow",
```

that is one entry: `NSStatusItem Preferred Position Item-0NSWindow Frame SettingsSheetWindow`.

`scanForUnknownPreferencesInDictionary` then warns that both real keys are unknown. It does not delete them.

The status-item hide/show bug in #906 is separate and already fixed by not hiding the item. This comma was omitted when the key was added in [`1b29a9c0ea`](https://github.com/Tunnelblick/Tunnelblick/commit/1b29a9c0ea566dd671a5c5cdab5d99ad953157a9) (2026-07-08).

## Change

Add the missing comma so both keys are separate allowlist entries.

## Validation

A small C program confirms adjacent literals concatenate (`concat_len=72`, the two names meet at `0N`) and that a two-element array keeps them separate. This repo has no first-party unit harness.

Ref #906

### DIFF
--- a/tunnelblick/defines.h
+++ b/tunnelblick/defines.h
@@ -1057,7 +1057,7 @@ name = newValue;                                    \
 @"updateCheckInterval",	\
 @"updateFeedURL",	\
 \
-@"NSStatusItem Preferred Position Item-0" \
+@"NSStatusItem Preferred Position Item-0",	\
 @"NSWindow Frame SettingsSheetWindow",	\
 @"NSWindow Frame ConnectingWindow",	\
 @"NSWindow Frame SUStatusFrame",	\


jkbullard:
Thanks!

PR #914, opened 2026-08-24
fix: start disconnect before waiting when a config is removed

SebTardif:
Deleting or replacing a connected configuration queued the disconnect on the main thread and then waited on that same thread, so OpenVPN was never signaled and the UI never returned.

## Problem

`configurationsChanged` always runs on the main thread (`performSelectorOnMainThread`). `deleteExistingConfig` then did:

```
[myConnection performSelectorOnMainThread:@selector(startDisconnectingUserKnows:)
                               withObject:@YES
                            waitUntilDone:NO];
[myConnection waitUntilDisconnected];
```

`waitUntilDone:NO` queues the kill behind the current main-thread call. `waitUntilDisconnected` blocks that call until the OpenVPN pid is gone. The process is never killed, `waitUntilNoProcessWithID` times out every 5 seconds, and the outer loop never ends.

If `pid == 0` (still connecting, or waiting for network), the same loop never ends because the wait is skipped and `disconnectionComplete` stays false.

Public path: delete or move a connected `.tblk` while Tunnelblick is running.

This wait has been unbounded since at least 2014 ([`60865ae5f`](https://github.com/Tunnelblick/Tunnelblick/commit/60865ae5f)). Same class as #910.

## Change

- Run `startDisconnectingUserKnows:` with `waitUntilDone:YES` so the kill happens before the wait (on the main thread Apple invokes the selector immediately).
- If there is no OpenVPN pid, return immediately.
- Bound the remaining wait to 60 seconds and return `NO` on timeout.
- Show the "has been disconnected" alert only when the wait succeeds.

## Validation

Control-flow check: with `waitUntilDone:NO` on the main thread, `startDisconnectingUserKnows:` cannot run until `deleteExistingConfig` returns, which it never does. This repo has no first-party unit harness.

Ref #910

### DIFF
--- a/tunnelblick/MenuController.m
+++ b/tunnelblick/MenuController.m
@@ -2804,12 +2804,14 @@ static pthread_mutex_t configModifyMutex = PTHREAD_MUTEX_INITIALIZER;
     if (   myConnection
         && ( ! [[myConnection state] isEqualTo: @"EXITING"] )  ) {
         [myConnection addToLog: @"Disconnecting; user asked to delete the configuration"];
-        [myConnection performSelectorOnMainThread: @selector(startDisconnectingUserKnows:) withObject: @YES waitUntilDone: NO];
-        [myConnection waitUntilDisconnected];
+        [myConnection performSelectorOnMainThread: @selector(startDisconnectingUserKnows:) withObject: @YES waitUntilDone: YES];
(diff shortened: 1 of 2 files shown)


jkbullard:
Thanks!

Issue #814, opened 2024-08-13
Unable to set permissions error when up command given an executable


zaygraveyard:
**Describe the bug**
I get a "VPN Configuration Installation Error" when the config contains an up command with an executable instead of a script.

**To Reproduce**
1. Try to install a config with an `up` command with an executable, ex. `up "/bin/sh -c '/usr/sbin/ipconfig set $dev DHCP'"`
2. See error
```
Installation failed:

At line 141 of the OpenVPN configuration file: Unable to set permissions on '/private/var/folders/sk/fsvkfqk52z7d0mppbvjk30q00000gn/T/net.tunnelblick.tunnelblick-rIAcAJ/test.tblk/Contents/Resources/sh.sh'
```

**Expected behavior**
It works without errors.

**Additional context**
 - OS: macOS 14.5
 - Version `Tunnelblick 4.0.1 (build 5971)`

This issue seams related to #726 and the workaround suggested in https://github.com/Tunnelblick/Tunnelblick/issues/726#issuecomment-1149289926 works for this issue.

jkbullard:
@zaygraveyard - Thanks for reporting this.

The problem is that Tunnelblick expects the "command" to be a script, not an executable.

That will be fixed in the next beta release, with the following limitation: The executable must have an absolute path. So the following (your example) will install properly:

`up "/bin/sh -c '/usr/sbin/ipconfig set $dev DHCP'"`

But this still won't install properly, because sh is not an absolute path:

`up "sh -c '/usr/sbin/ipconfig set $dev DHCP'"`

(This issue will be closed automatically when the fix is committed publicly.)

zaygraveyard:
This limitation seems reasonable to me. 😄

Issue #726, opened 2022-06-02
Unable to load contents of cmd as UTF-8 when up command given an executable


zaygraveyard:
**Describe the bug**
I get a "VPN Configuration Installation Error" when the config contains an `up` command with an executable instead of a script.

**To Reproduce**
1. Try to install a config with an `up` command with an executable, ex. `up /bin/sh -c '/usr/sbin/ipconfig set $dev DHCP'`
2. See error
```
Installation failed:

At line 141 of the OpenVPN configuration file: Unable to load contents of /bin/sh as UTF-8
```

**Expected behavior**
It works without errors

**Desktop:**
 - OS: macOS 12.4
 - Version `Tunnelblick 3.8.7a (build 5770)`

**Additional context**
I believe that this should work as it is explicitly mentioned in the reference manual.

> --up cmd
>    Run command cmd after successful TUN/TAP device open (pre --user UID change).cmd consists of a path to script (**or executable program**), optionally followed by arguments.
Source: https://openvpn.net/community-resources/reference-manual-for-openvpn-2-4/

Without this functionality I'm forced to use an additional script that runs the same command, rendering the config impossible be self contained. Not a major issue but still.

jkbullard:
Thanks for your report. It is a bug.

As a workaround, I believe you can comment out the "up" (and "down" will have the same problem) to install the configuration, then edit it after it is installed to activate the "up" and "down" options.

zaygraveyard:
Thank you for the workaround 😄 
FWIW I had to change the up command to `up "/bin/sh -c '/usr/sbin/ipconfig set $dev DHCP'"` (note the double quotes)

jkbullard:
Fixed in Tunnelblick 4.0.0beta03.

jkbullard:
It probably won't be fixed in Tunnelblick 3. No idea when 4.0.0 will be the stable version, it depends on further testing.

PR #915, opened 2026-08-24
fix: copy non-UTF-8 config files as binary

SebTardif:
A non-UTF-8 `.sh` / `.key` / `.pem` / `.crt` / `.crl` was logged as a UTF-8 failure and then treated as a successful copy, so the dest file was never written.

## Problem

`duplicateFileFrom:toPath:` returns nil for success. After #726 ([`0f139357b`](https://github.com/Tunnelblick/Tunnelblick/commit/0f139357b4afb985e1c73912ac5133bb0eb4b7fd)), a UTF-8 decode failure logged the error and `return nil` without writing the dest file. Callers treated that as success and then `checkSetPermissions` ran on a missing dest (`#814`).

The #726 intent was to accept a binary `up` executable (for example `/bin/sh`). That needs a binary copy, not a silent skip.

## Change

On UTF-8 decode failure, keep the existing log line and fall through to the existing `tbCopyPath` branch. Valid UTF-8 files still get CR conversion as before.

## Validation

The success contract is `return nil`. The old UTF-8-failure path returned nil without creating `target`. The new path always calls `tbCopyPath` when decode fails. This repo has no first-party unit harness.

Ref #726
Ref #814

### DIFF
--- a/tunnelblick/ConfigurationConverter.m
+++ b/tunnelblick/ConfigurationConverter.m
@@ -556,12 +556,7 @@ TBSYNTHESIZE_OBJECT_GET(retain, NSString *, nameForErrorMessages)
 		}
 		
 		NSMutableString * contents = [[[NSMutableString alloc] initWithData: data encoding: NSUTF8StringEncoding] autorelease];
-		if (  contents == nil  ) {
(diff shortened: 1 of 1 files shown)


jkbullard:
Thanks!

PR #919, opened 2026-08-26
fix: public robustness for updater, installer, and GUI leftovers

SebTardif:
These changes fix the remaining public audit findings on `main`. They do not include the two private GHSA items.

## Problem

Several leftover GUI, installer, and updater paths can hang, leak, overwrite user files, or treat a failed install as success.

- `runTunnelblickd` stores `write`/`read` in a `size_t`. `n < 0` is then dead. A failed socket I/O wraps the pointer or tries to allocate `SIZE_MAX` bytes.
- `usingOurEasyRsa` compares a raw `~/...` preference to an expanded path, so Tunnelblick can overwrite a user easy-rsa tree.
- The OpenVPN safe-option list is missing a comma, so `ifconfig-pool` and `ifconfig-push-constraint` become one token. Installing a config that uses either option gets the “TAKE COMPLETE CONTROL OF YOUR COMPUTER” prompt.
- `INSTALLER_INSTALL_FORCED_PREFERENCES` deletes `forced-preferences.plist` before the copy. A failed copy leaves no forced prefs.
- Welcome-page link clicks open the system browser but never call `[listener ignore]`, so the in-process WebView can still navigate.
- Safe-config install failures increment `nSafeErrors` but not `nTotalErrors`, so the installer can report success.
- Offline update retries leak `NSURLConnection` objects and ignore `stopDownload` while waiting to retry.
- The phase-3 updater helper waits for the first process named Tunnelblick, with no deadline.
- `TBButton` / `TBPopUpButton` `dealloc` skips `tracker`, so a help popup can leak the button.

## Change

- `sharedRoutines.m`: `ssize_t n` so write/read errors take `error1`.
- `easyRsa.m`: expand `easy-rsaPath` with `stringByExpandingTildeInPath` before the compare.
- `defines.h`: comma after `@"ifconfig-pool"`.
- `installer.m`: copy to a `.new` file, set owner/mode, `rename` over the live plist.
- `WelcomeController.m`: `[listener ignore]` after `openURL:`.
- `TBButton.m` / `TBPopUpButton.m`: release `tracker` in `dealloc`.
- `ConfigurationManager.m`: add `nSafeErrors` to `nTotalErrors`.
- `TBDownloader.m`: release the previous connection on retry; `stopDownload` always invalidates the retry timer; `dealloc` invalidates the timer.
- `TunnelblickUpdateHelper.m`: wait with `kill(pid, 0)` for 60s. If the process is still running, log and `errorExit()`. Do not SIGKILL.

## Validation

Tunnelblick has no first-party unit harness and no GitHub Actions on this repo.

Host check for the `size_t` compare:

```
size_t n<0=0 ssize_t s<0=1
```

C string concatenation of the two option names produces `ifconfig-poolifconfig-push-constraint` without the comma.

Same patterns as merged #912 (safelist comma) and #917 (do not delete forced prefs first).

## Origin

- `size_t n`: [`6961b8778`](https://github.com/Tunnelblick/Tunnelblick/commit/6961b8778) (2023-05-31, “fix warnings”).
- Offline retry: [`67492ce4e`](https://github.com/Tunnelblick/Tunnelblick/commit/67492ce4e) (2025-02-14).
- Welcome policy: [`4fb4bf6e7`](https://github.com/Tunnelblick/Tunnelblick/commit/4fb4bf6e7) (2012-09-12).
- Missing safelist comma: present on current `main`.

Related in-repo: #912, #917, #910, #916.

### DIFF
--- a/tunnelblick/ConfigurationManager.m
+++ b/tunnelblick/ConfigurationManager.m
@@ -2631,7 +2631,7 @@ in: (NSString *) sharedOrPrivate {
 
     // Construct and display a window with the results of the uninstalls/replacements/installs
 
-    NSUInteger nTotalErrors = nUninstallErrors + nInstallErrors + nReplaceErrors + nUpdateErrors;
+    NSUInteger nTotalErrors = nUninstallErrors + nInstallErrors + nReplaceErrors + nUpdateErrors + nSafeErrors;
 
     if (   (nTotalErrors != 0)
         || ( ! skipResultMsg )  ) {
--- a/tunnelblick/LeftNavItem.m
+++ b/tunnelblick/LeftNavItem.m
@@ -39,7 +39,7 @@ static LeftNavItem * rootItem = nil;
 	
     [displayName             release]; displayName             = nil;
     [nameToShowInOutlineView release]; nameToShowInOutlineView = nil;
-    [parent                  release]; parent                  = nil;
+    parent                                                         = nil;
     [children                release]; children                = nil;
 	
     [super dealloc];
@@ -58,7 +58,14 @@ static LeftNavItem * rootItem = nil;
 
 TBSYNTHESIZE_OBJECT(retain, NSString *,       displayName,             setDisplayName             )
 TBSYNTHESIZE_OBJECT(retain, NSString *,       nameToShowInOutlineView, setNameToShowInOutlineView )
-TBSYNTHESIZE_OBJECT(retain, LeftNavItem *,    parent,                  setParent                  )
 TBSYNTHESIZE_OBJECT(retain, NSMutableArray *, children,                setChildren                )
+
+-(LeftNavItem *) parent {
+    return parent;
+}
+
+-(void) setParent: (LeftNavItem *) newValue {
+    parent = newValue;
+}
 
 @end
--- a/tunnelblick/LeftNavItem.m
+++ b/tunnelblick/LeftNavItem.m
@@ -39,7 +39,7 @@ static LeftNavItem * rootItem = nil;
 	
     [displayName             release]; displayName             = nil;
     [nameToShowInOutlineView release]; nameToShowInOutlineView = nil;
-    parent                                                         = nil;
+    [parent                  release]; parent                  = nil;
     [children                release]; children                = nil;
 	
     [super dealloc];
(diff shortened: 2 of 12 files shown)


jkbullard:
@SebTardif - **Thank you very much for this PR (and for your other PRs and GHSAs)!**

Here are my comments on the changes in this patch:

<hr>

**ConfigurationManager.m, defines.h, easyRsa.m, installer.m, sharedRoutines.m,TBButton.m, TBDownloader.m, TBPopUpButton.m, and WelcomeController.m**: all changes look good.

<hr>

**LeftNavItem.m and Tracker.m**: These changes should not be made. They apparently assume Automatic Reference Counting ("ARC") is active, but Tunnelblick does not use ARC.

**Please remove the changes.**

<hr>

**TunnelblickUpdateHelper.m**: I think the intent of the patch is to fix a hang if the invoking user's Tunnelblick process does not terminate within 60 seconds. The patch does that by logging the timeout and invoking errorExit() to terminate the Tunnelblick update.

But the changes to waitUntilNoProcessWithProcessID() also affect the behavior of waitUntilNoProcessWithName(), which is used to wait for other users' Tunnelblick applications and the tunnelblickd daemon process to terminate. For these, if the process does not terminate before new timeout period, the process is SIGKILLed, and the update process continues without indicating any error even if the process continues after a further 2 seconds. SIGKILL should not be sent to the process, and errorExit() should be called if a timeout occurs and the process is still running.

**Please remove the SIGKILL code, and log and errorExit() if a timeout occurs with a process still running after the original timeout period even if sendSIGTERM was TRUE.**

There is also another behavior change for those processes: if the process terminates and macOS quickly reassigns the process's pid to a new, unrelated process, waitUntilNoProcessWithProcessID() will, if sendSIGTERM is TRUE, SIGKILL that (unrelated!) process. Removing the SIGKILL code will fix that problem. In the same circumstances, the original code will SIGTERM a Tunnelblick or tunnelblickd process and then hang forever (as if the original process did not terminate). That's slightly better than SIGKILLing an unrelated process, but it's also bad!

The use of process ID is problematic, and should be looked at further. As I recall, when I wrote the original code I had a reference that a macOS process's process ID is not reused for some period of time after the process terminates, but I can't find anything about that now.

SebTardif:
@jkbullard

> Thank you very much for this PR (and for your other PRs and GHSAs)!

Thanks for the review.

> **LeftNavItem.m and Tracker.m**: These changes should not be made. They apparently assume Automatic Reference Counting ("ARC") is active, but Tunnelblick does not use ARC.
>
> **Please remove the changes.**

Reverted both files. The rest of the PR is unchanged, including the `TBButton` / `TBPopUpButton` `tracker` releases in `dealloc`.

> **Please remove the SIGKILL code, and log and errorExit() if a timeout occurs with a process still running after the original timeout period even if sendSIGTERM was TRUE.**

Done in `waitUntilNoProcessWithProcessID()`. After the 60 second wait, if `kill(pid, 0)` still succeeds, it logs and calls `errorExit()`. There is no SIGKILL path, including when `sendSIGTERM` is true. That also avoids sending SIGKILL to a reused pid.

